PR #86, opened 2019-01-20
 Support for external fonts

belek666:
This commit adds:
- Loading "sh2e\font\font000.tga" image which contains sets of two fonts and patching game to use it. If option `UseCustomFonts = 0` or image file is missing/format is wrong patches won't be loaded.
- Loading "sh2e\font\fontwdata.bin" as width data for the first 224 chars, for two sets of fonts (optional). The file contains width values in byte size. If it's missing default game's values will be used.
- Reading int values from *.ini file to get font's settings.

### DIFF
--- a/Common/Settings.cpp
+++ b/Common/Settings.cpp
@@ -28,6 +28,11 @@
 
 VISIT_BOOL_SETTINGS(SET_BOOL_DEFAULTS);
 
+#define SET_INT_DEFAULTS(name, value) \
+	int name = value;
+
+VISIT_INT_SETTINGS(SET_INT_DEFAULTS);
+
 // Forward declarations
 bool SetValue(char* name);
 bool IsValidSettings(char* name, char* value);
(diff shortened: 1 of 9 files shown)


Polymega:
@elishacloud 
Belek666 has now given us the ability to add high resolution fonts to the game (example [here](https://camo.githubusercontent.com/c2c90e766eea8927cca2a500074fe56efdfa33bd/68747470733a2f2f692e696d6775722e636f6d2f644f36305566452e706e67) and [here](https://camo.githubusercontent.com/f8f1b4d5dc03159609add90d4907e830244fce27/68747470733a2f2f692e696d6775722e636f6d2f62376235496d662e706e67))! A major improvement to the very low-res font we currently have! 

Aero_ and I were able to play test Belek666's custom font mod and it works great across all three executables (1.0, 1.1, & DC). As mentioned in his pull request, there will be a new section in the `d3d8.ini` dedicated to adjusting the custom font, along with two external files: 

```
\sh2e\font\fontwdata.bin - contains optional font width data
\sh2e\font\font000.tga - image file for the new, higher res font
```

I wanted to re-mention that part because the .tga file is 15 MB in size. Would you like to include the .tga file as part of the download package for the SH2 Enhancement mod, or would you prefer me to include it as part of the Enhanced Edition Essential Files download package due to its size?

Thank you again Belek666 for all your hard work with this improvement!

elishacloud:
@belek666, nice work!

@Polymega, please add them to the Enhanced Edition Essential Files download package, since they are under the sh2e mod folder.

Issue #88, opened 2019-01-23
Issues with Main and In-Game menus


teh53m1chr157:
Whenever I click on the controls section of the Options menu, it takes around 20 seconds to load the next screen as well as when I exit said menu screen back to Options and subsequently, back to the Main menu. This also happens when opening the Inventory menu or when examining objects, such as Mary's letter at the beginning.

Edit: Apologies for the double-post, my VPN caused some lag.

teh53m1chr157:
Ah! Perfect! Thanks so much. I was hitting a brick wall Googling the subject, lol.

teh53m1chr157:
I am however having trouble getting the D-Pad to function as movement through XInput. The suggestion on the official page is not working. It should be noted that I had previously set the configuration without that option. Do I need to delete dlls before trying again?

Polymega:
Glad to hear! May I ask what your GPU and display resolution is? I'd like to see if there's a common denominator for this. 

Also, I've been dragging my feet on it, but I plan to update ReShade and, if tests are good, share an updated version on the Installation Guide. Maybe the updated version will allow you to use ReShade again. I'll ping you once I have those files ready down the line. 

> It should be noted that I had previously set the configuration without that option. Do I need to delete dlls before trying again?

All you should need to do is open the Xinput Plus software again, set sh2pc.exe as the target file, and follow the steps on the guide to make the d-pad work as a movement pad. You'll need to apply these new changes through the Xinput Plus software (thus, having it overwrite its previous .dll files it makes in the game's directory). 

And, as always, you'll need to relaunch the game any time you adjust external settings to see the changes take effect.

Polymega:
@teh53m1chr157 
Do me a favor and try out this new ReShade package. It's updated to v 4.1.0. Let me know if that resolves the lag/stuttering while still being able to use ReShade.

1. Delete the following files/folders from the game's installation directory, if you haven't already done so: 
   - `\reshade-shaders\`
   - `d3d9.dll`
   - `d3d9.ini`
   - `d3d9.log`
   - `ReShade Filter - SH2 PC PS2.ini`
2. [Download the new ReShade package. ](http://www.enhanced.townofsilenthill.com/SH2/files/SH2PC_ReShade_4.1.0_DirectX9.zip)
   - Place the files in your game's installation directory
3. Ensure `d3d8to9 =` is set to `1` in the `d3d8.ini` file.

teh53m1chr157:
[...]

PR #99, opened 2019-03-12
Replaced DirectInput force feedback with XInput vibration

CookiePLMonster:
This feature was made as kindly requested by Ratiocinator.

Reportedly, XInput Plus vibration broke completely at some point, so SH2 was left with no controller rumble at all. I spent some time trying to figure out why XI Plus broke, but with no success. Eventually, I noticed how simple SH2's rumble is and decided to just reimplement it with XInput.

A few notes:
1. Wrapper for `IDirectInputEffect` is extremely primitive and game specific. It literally is tailored for how SH2 is using vibration, with nothing else implemented.
2. I did **not** link the lib against DX SDK, so it is creating a dependency on `xinput9_1_0.dll`. This DLL ships by default with all Windows since WinXP, so no dependency on DirectX Redistributables is created.

If you have any doubts or suggestions about either the code or hooking style, please don't hesitate to ask.

### DIFF
--- a/Common/Settings.h
+++ b/Common/Settings.h
@@ -36,7 +36,8 @@
 	visit(WidescreenFix, true) \
 	visit(WndModeBorder, true) \
 	visit(UseCustomFonts, false) \
-	visit(DisableEnlargedText, false)
+	visit(DisableEnlargedText, false) \
+	visit(XInputVibration, false)
 
 #define VISIT_INT_SETTINGS(visit) \
 	visit(CustomFontCol, 100) \
@@ -46,7 +47,8 @@
 	visit(NormalFontWidth, 20) \
 	visit(NormalFontHeight, 30) \
 	visit(SmallFontWidth, 14) \
-	visit(SmallFontHeight, 24)
+	visit(SmallFontHeight, 24) \
+	visit(PadNumber, 0)
 
 // Configurable setting defaults
 #define DECLARE_BOOL_SETTINGS(name, unused) \
--- a/Common/Settings.ini
+++ b/Common/Settings.ini
@@ -5,6 +5,7 @@ AutoUpdateModule = 1 // Checks for updated versions of the SH2 Enhancement modul
 d3d8to9 = 1 // Causes Silent Hill 2 to use DirectX 9.
 UseCustomModFolder = 1 // Uses files in the 'sh2e' folder rather than the 'data' folder.
 WidescreenFix = 1 // Enables WidescreenFixesPack and sh2proxy module. See options below.
+XInputVibration = 1 // Replaces DirectInput based vibration with XInput based vibration.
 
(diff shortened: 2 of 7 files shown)


CookiePLMonster:
I pushed a small change on Ratiocinator's request - now vibration is submitted to **both** DirectInput and XInput, so DInput pads can still function out of the box, without having to disable this option in INI.

If this ever causes problems, killing DInput vibration is trivial (`Release()` the old pointer instead of copying it, done).

Polymega:
Thanks for all that you're doing here, @CookiePLMonster ! 

@elishacloud unrelated, but, when you get to merging this PR, may I also suggest enabling borderless window mode by default for people who download the project? [AMD cards experience sluggish behavior with the `FastTransitions` mod when played in fullscreen mode](https://github.com/elishacloud/Silent-Hill-2-Enhancements/issues/88#issuecomment-456641967), so to play things safe for first-timers, we could have it windowed but fill their whole screens. I reckon no one would complain about that, let alone be none the wiser.

AeroWidescreen:
@Polymega That's one of the few things you should be able to change with no hassle. Edit the files then create a pull request when you're done.

Defaults: https://github.com/elishacloud/Silent-Hill-2-Enhancements/blob/master/Common/Settings.h
Ini: https://github.com/elishacloud/Silent-Hill-2-Enhancements/blob/master/Common/Settings.ini

You could go a step further and build the project using Visual Studio to test the results. I'm not saying you should do any of this now, but it's something you could experiment with in the future. 👍

Polymega:
Thanks for the tip, Aero_. 

I could do this for simple tasks such as recommending/adjusting default .ini settings (although I wouldn't have thought to also edit settings.h, only settings.ini, as I'm so unfamiliar with programming!). 

I wouldn't dare attempt anything more complicated than that, as I'm not kidding when I say I'm not a programmer to save my life.

elishacloud:
@CookiePLMonster, for some reason when I compile with this the game hangs if `XInputVibration` is enabled.  I tried installing [XInput Plus](https://sites.google.com/site/0dd14lab/xinput-plus) and that did not make any difference.

Here is a compiled version in case anyone else wants to test it: [d3d8.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/3011440/d3d8.zip)

Polymega:
Can confirm on my end, too, from an earlier test build Silent sent me. If the fix is enabled but no controller is plugged in on game launch it crashes. Works great if a pad is plugged in prior to launch, though.

CookiePLMonster:
D'oh! I totally overlooked the case where a DInput gamepad doesn't initialize at all - fixed now.

elishacloud:
Thanks @CookiePLMonster, the update looks good.

PR #100, opened 2019-03-14
Set EnableWndMode by default in .ini

Polymega:
(Cautiously dipping my toes into a pull request; first time doing this.)

Simply enables `EnableWndMode` by default for users who download the project. This fixes the issue of the `FastTransitions` mod having noticeable slow-down for AMD users when played in fullscreen mode.

### DIFF
--- a/Common/Settings.h
+++ b/Common/Settings.h
@@ -15,7 +15,7 @@
 	visit(FixMissingWallChunks, true) \
 	visit(Fog2DFix, true) \
 	visit(FogParameterFix, true) \
-	visit(FullscreenWndMode, false) \
+	visit(FullscreenWndMode, true) \
 	visit(HalogenLightFix, true) \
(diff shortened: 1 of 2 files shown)


Keith94:
Not `FullscreenWndMode = 1`, too?

elishacloud:
I agree with Keith94 on this one.  Enabling `FullscreenWndMode` by default is a good idea here incase someone disables `WidescreenFix`.  This will still allow it to run in fullscreen even in lower resolutions.

Polymega:
Yes, and thank you both for advice on that. Because of the continuous amounts of play testing I do for the project, I leave that feature off to make it easier for me to alt + tab out of the game and, by having it off for so long now, it completely skipped my mind to also have that on for first-timers, as well. 

Sorry if I'm butchering commit optimizations with GitHub as well; this is new territory for me and am experimenting around with it.

PR #108, opened 2019-04-16
Added DPadMovementFix

CookiePLMonster:
This feature was made as kindly requested by Ratiocinator.

This PR adds a `DPadMovementFix` setting. When enabled, following changes are made:

1. DPad buttons mirror analog stick behaviour. This means you are able to walk around and navigate menus using a DPad **and** analog stick.
2. Corrects hotplugging behaviour a bit. Previously, if the game was launched with a controller plugged in, then it was plugged out while performing some action (eg. having a button pressed or analog stick pushed) input would "stick", and would not go away until the game is restarted or controller is plugged back in.

It's been tested and proven to work with and without controllers and on 1.0, 1.1 and DC EXEs.

### DIFF
--- a/Common/Settings.h
+++ b/Common/Settings.h
@@ -40,6 +40,7 @@
 	visit(UseCustomFonts, true) \
 	visit(DisableEnlargedText, true) \
 	visit(XInputVibration, true) \
+	visit(DPadMovementFix, true) \
 	visit(UseCustomExeStr, true) \
 	visit(UnlockJapLang, false)
 
--- a/Common/Settings.ini
+++ b/Common/Settings.ini
@@ -6,6 +6,7 @@ d3d8to9 = 1 // Causes Silent Hill 2 to use DirectX 9.
 UseCustomModFolder = 1 // Uses files in the 'sh2e' folder rather than the 'data' folder.
 WidescreenFix = 1 // Enables WidescreenFixesPack and sh2proxy module. See options below.
 XInputVibration = 1 // Replaces DirectInput based vibration with XInput based vibration.
+DPadMovementFix = 1 // Allow movement with DPad
 UseCustomExeStr = 1 // Loads "exe_str.txt" file located in the "sh2e\text" folder as built-in strings. Restores the language selection in the Options menu.
 UnlockJapLang = 0 // Unlocks Japanese language in the Options menu. It is recommended to keep this disabled for now, as not all Japanese characters are currently supported.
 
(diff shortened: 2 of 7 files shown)


Keith94:
>This means you are able to walk around and navigate menus using a DPad and analog stick.

CookiePLMonster:
I believe so, yes. That's how it was pitched to me at least.

Polymega:
Yes, after this the only remaining setback for the PC version is that you can't use the right joystick to move the Search Camera; it's tied to the left joystick (and now d-pad) on the PC build which means you must fully stop moving James to move the Search Camera.

Polymega:
Hey @seagull !

Our friend Silent (@CookiePLMonster) has created native D-pad support for DInput gamepads. Your DualShock 4 controller's D-pad will work out of the box now. :)

seagull:
Thanks for the tip! Great work, CookiePLMonster!

AeroWidescreen:
@CookiePLMonster. Very happy to see dpad support added. That always felt like a strange exclusion. Thank you for your contributions, big fan of your SilentPatches too. 😁 

@Polymega Do you know if this has any effect for XInput controllers? I know we can use XInput Plus to achieve the same thing, but it would be nice to not have to rely on that as much.

Polymega:
Yes, I tested this with the following controllers:  

Xbox 360 (wired), Xbox One (wireless via USB dongle), Logitech RumblePad 2, and DualShock 4. 

Silent's DPadMovementFix works on all these controllers both with and without XInput Plus. However, XInput Plus is still required for XInput controllers (such as the Xbox controllers) because the game won't recognize its triggers without it. The game also didn't recognize the D-pad for XInput controllers but that has since been fixed thanks to Silent's work here. 

Additionally, Silent's restored vibration fix works with XInput pads without XInput Plus as well. In short: We only use XInput Plus to assigns functions to an XInput pad's triggers now.

The only potential headache we may receive in terms of troubleshooting is if someone downloads the XInput Plus files then proceeds to use a DInput gamepad, as XInput Plus disables vibration for DInput pads. A user would either need to remove the XInput Plus files to restore vibration on DInput pads, or set `EnableX2Dinput = False` in the `XinputPlus.ini` file to restore the rumble. They'd have to set that back to `True` if they ever decide to use a XInput pad with XInput Plus in the future in order for the triggers and et al to work.

elishacloud:
Nice work!  Thanks @CookiePLMonster!

PR #140, opened 2019-06-27
Main Menu/Options fixes

belek666:
This commit adds:
- resolution lock and displaying current resolution in Advanced Option menu
- fix for mouse hit-boxes in Main Menu (version 1.1 only)
- per language 'start01x.tex' file (graphic Main Menu)
- fix text and mouse hit-boxes misalignments in Game Option menu
- make "Born From a Wish" horizontal position the same as "Letter From Silent Heaven" in New Game menu
- update "Now loading" text in Main Menu according to language


Polymega:
Thank you again, Belek!! 

Elisha was working on an update to fix constant lag/stuttering that resulted from Windows Updates (rundll32.exe and gameux.dll issues). As more people update their Windows, that issue would become a bigger problem as time went on, so Elisha just released a quick hot fix update for the project to alleviate that problem.

When we release the next build you're amazing work will be a part of that update!

belek666:
@Polymega Thanks for your help.
@elishacloud Should I update build number? To which one?

elishacloud:
@belek666, thanks for the updates!.  Don't worry about the build number.  I will update that.

However, I did find some issues with this update.  When using the v1.0 binary I get an error message in the log file:
```text
16060 21:37:30.225 Enabling Resolution Lock...
16060 21:37:30.228 UpdateMainMenuFix Error: failed to find memory address!
16060 21:37:30.229 Enabling Main Menu title selection by language...
```

When using the DC binary I get two error messages in the log file:
```text
20392 21:28:06.666 Enabling Resolution Lock...
20392 21:28:06.670 UpdateMainMenuFix Error: failed to find memory address!
20392 21:28:06.678 UpdateMainMenuTitlePerLang Error: failed to find memory address!
```

Can you fix these errors?

Polymega:
We'll get Belek's word on this but I believe I know what's going on with the errors: 

---------------

The `MainMenuFix` fixes the mouse misalignment issues for the main menu's text hit boxes on 1.1 only. (The hitboxes are inherently fine on both 1.0 and DC). So this error would probably only show up in 1.0 and DC.

---------------

The `MainMenuTitlePerLang` allows us to use custom, translated `start01.tex` files (main menu font images), saved as `start01*.tex`, for 1.0 and 1.1. Here's a video example: https://twitter.com/WhereAllBegins/status/1146238025912999936

For the DC binary, that version uses live text to populate the main menu, so it never uses/calls the `start01*.tex` images. The error you see for that on DC is probably the reason why.

----------------

belek666:
@elishacloud @Polymega  is right. They are fixes for specified exe version.

elishacloud:
Thanks for the work on this @belek666!   I merged the changes.  I also made a few updates, mainly to the `LockResolution` fix:
1. I used the actual resolution given to Direct3D8 by the game, rather than the value in the ini file.
2. I made it so that this is only enabled if the WSF module gets loaded.  Otherwise you cannot change the resolution when WSF is disabled, which used to work.

As far as the error messages, I plan to add detection of the binary version (1.0, 1.1 or DC) to the project later.  Once I do that I can enable/disable the appropriate features based on the binary version.  This will remove the error messages in the log file.  Also it would be good to log the binary version to assist in future troubleshooting.

PR #154, opened 2019-07-13
Merge Soft Shadows to master

elishacloud:


umbrellacorp53:
Does this mean the soft shadows have been implemented into the main updated d3d8.dll?

elishacloud:
The code is checked in, but we have not released an update yet.

Here is a build you can use for soft shadows: [d3d8.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/3389246/d3d8.zip)

_Note: you will need to add `EnableSoftShadows = 1` to the ini file to enable soft shadows._

umbrellacorp53:
Perfect. Is this any different than frozenfish's files? His files only worked with the EN 1.0 .exe. Does this work with the director's cut?

Also, I tried with self shadows and soft shadows. Some slightly odd visuals but mostly **awesome!**

elishacloud:
Yes, this will work with all versions of Silent Hill 2.

PR #156, opened 2019-07-14
Scale shadow dimensions 50% before blurring

FrozenFish24:
75% less pixels to process and allows a similar level of softness with only 4 blur passes vs 16.

### DIFF
--- a/Wrappers/d3d8/IDirect3DDevice8.cpp
+++ b/Wrappers/d3d8/IDirect3DDevice8.cpp
@@ -1338,11 +1338,16 @@ HRESULT m_IDirect3DDevice8::GetInfo(THIS_ DWORD DevInfoID, void* pDevInfoStruct,
 
 HRESULT m_IDirect3DDevice8::DrawSoftShadows()
 {
+	// Variables for soft shadows
+	DWORD SHADOW_OPACITY = 128;
+	int SHADOW_DIVISOR = 2;
+	int BLUR_PASSES = 4;
+
 	float screenW = (float)BufferWidth;
 	float screenH = (float)BufferHeight;
 
-	// Varables for soft shadows
-	DWORD SHADOW_OPACITY = 128;
+	float screenWs = (float)BufferWidth / SHADOW_DIVISOR;
+	float screenHs = (float)BufferHeight / SHADOW_DIVISOR;
 
 	// Original geometry vanilla game uses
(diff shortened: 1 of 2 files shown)


Polymega:
Little bug to report: When soft shadows is enabled (`EnableSoftShadows = 1`) the game screen will turn white and stop responding if you alt + tab out of the game and open another application that provokes the UAC prompt on Windows 7. It might do the same for Windows 10 but I don't have the means to test it.

elishacloud:
I have not tested it yet, but the most likely cause of this is because the game called the Reset() function and the textures used for the soft shadows have not been destroyed causing the Reset to fail.

The fix for this is to track the Reset() function and free the texture before the Reset() has been called.  Then we will automatically recreate them next time we run the soft shadows function.

Anyways that is a separate issue and not related to this pull request.  I will merge this and put in a fix for the other item soon.

Here are the details from [the documentation](https://docs.microsoft.com/en-us/windows/win32/api/d3d9/nf-d3d9-idirect3ddevice9-reset):

> Calling IDirect3DDevice9::Reset causes all texture memory surfaces to be lost, managed textures to be flushed from video memory, and all state information to be lost. Before calling the IDirect3DDevice9::Reset method for a device, an application should release any explicit render targets, depth stencil surfaces, additional swap chains, state blocks, and D3DPOOL_DEFAULT resources associated with the device.

_Note: The textures used for soft shadows are created as D3DPOOL_DEFAULT resources._

Polymega:
Thank you both!

elishacloud:
@Polymega, I fixed the alt+tab issue in the latest [check in](https://github.com/elishacloud/Silent-Hill-2-Enhancements/commit/b44930f0ae246edf58f5b5721c2aefc735d06758).  There was an issue with the ref count of some textures and surfaces.  I added the proper release to fix ref count issues.  Also made sure we handled Reset() correctly.

Here is the build that has the fix:  [d3d8.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/3395612/d3d8.zip)

Polymega:
Thanks so much, Elisha. This will save me full game restarts whenever I launch Cheat Engine or the like.

FrozenFish24:
Nice, @elishacloud! I wouldn't have had about that one.

elishacloud:
@FrozenFish24, that is what is so great about working with a team!  I would not have figured out the soft shadows fix!

PR #164, opened 2019-08-01
Commit adds:

belek666:
- Fix drawing line when using '0xFE' byte in '.mes' files
- Option to change space between letters

### DIFF
--- a/Common/Settings.h
+++ b/Common/Settings.h
@@ -18,6 +18,7 @@
 	visit(EnableSelfShadows, false) \
 	visit(EnableSoftShadows, false) \
 	visit(EnableWndMode, true) \
+	visit(FixDrawingTextLine, true) \
 	visit(FixHangOnEsc, true) \
 	visit(FixMissingWallChunks, true) \
 	visit(Fog2DFix, true) \
@@ -60,6 +61,7 @@
 	visit(CustomFontRow, 14) \
 	visit(CustomFontCharWidth, 20) \
 	visit(CustomFontCharHeight, 32) \
+	visit(LetterSpacing, 2) \
 	visit(NormalFontWidth, 20) \
 	visit(NormalFontHeight, 30) \
 	visit(PadNumber, 0) \
--- a/Common/Settings.ini
+++ b/Common/Settings.ini
@@ -89,6 +89,8 @@ MainMenuTitlePerLang = 1
 UseCustomExeStr = 1
 ; Unlocks Japanese language in the Options menu. It is recommended to keep this disabled for now, as not all Japanese characters are currently supported.
 UnlockJapLang = 0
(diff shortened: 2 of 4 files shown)


belek666:
@elishacloud 
I'm not completely sure about solution for fixing drawing line. Game may use line primitives somewhere else. From my tests it looks ok. I tried to change only parameters in function responsible for drawing line. But game starts crashing before launch. It collide with 'Fix2D' option. Here is a code that I was trying to use:
```
constexpr BYTE LineSearchBytes[] = { 0xD8, 0x44, 0x24, 0x18, 0xD9, 0x9C, 0x24, 0x90, 0x00, 0x00, 0x00, 0xD9, 0x9C, 0x24, 0x94, 0x00, 0x00, 0x00 };

void FixLine()
{
	void *DLineAddr = CheckMultiMemoryAddress((void*)0x00482125, (void*)0x004823C5, (void*)0x004825D5, (void*)LineSearchBytes, sizeof(LineSearchBytes));

	// Search for address
	if (!DLineAddr) {
		Logging::Log() << __FUNCTION__ << " searching for memory address!";
		DLineAddr = GetAddressOfData(LineSearchBytes, sizeof(LineSearchBytes), 1, 0x00482000, 1800);
	}

	// Checking address pointer
	if (!DLineAddr) {
		Logging::Log() << __FUNCTION__ << " Error: failed to find memory address!";
		return;
	}

	BYTE val = 88;
	UpdateMemoryAddress((void *)((BYTE*)DLineAddr + 0x20), (void *)&val, 1);
	val = 2;
	UpdateMemoryAddress((void *)((BYTE*)DLineAddr + 0x23), (void *)&val, 1);
	val = 5;
	UpdateMemoryAddress((void *)((BYTE*)DLineAddr + 0x25), (void *)&val, 1);
}
```

elishacloud:
@belek666, Thanks for your work here! Sorry it took so long to commit this.  I'm not too sure about the Fix2D issues.  That is not part of my code.  You can see the code for that [here](https://github.com/ThirteenAG/WidescreenFixesPack/blob/4107ab19a52bf8ddc97cc823f5de339b0ef6499f/source/SilentHill2.WidescreenFix/dllmain.cpp#L77-L338).

Polymega:
@AeroWidescreen might be able to look over Belek's original, universal solution in regards to the `Fix2D` conflict. (I believe Aero_ is the creator of `Fix2D`).

But if the alternate solution is good then that's what matters! Thank you again Belek, and thanks for committing this Elisha; I'll playtest this tomorrow with the Lying Figure Vehicle fix (both are added to that test build?).

elishacloud:
> both are added to that test build?

Yes, both are added to the test build.  I also added the Windows XP compatibly to the build.  This _should_ allow it to run on Windows XP, at least until we update to Visual Studio 2019.

PR #172, opened 2019-09-22
Option to change size of space

belek666:
Simple option to change size of space.


elishacloud:
Nice work again @belek666!

PR #178, opened 2019-10-11
Adds 'FixTownWestGateEvent' option that changes James' commentary abo…

belek666:
…ut the back alley Heaven's Night gate at night to properly reflect the gate's status.


Polymega:
Forgot to mention on the other thread Elisha but this works perfect across all binaries!

PR #181, opened 2019-10-23
Make shadows act more like Xbox

FrozenFish24:
This pull request allows us to easily exclude dynamic objects like James from receiving self shadows.

Basic idea:
1. Draw map and character geometry as usual, anything with `STENCILOP_REPLACE` will draw it's silhouette to the stencil buffer, anything with `STENCILOP_ZERO` with mask it's silhouette from the stencil buffer
2. Draw off the stencil buffer to the color buffer of a new texture, we now have a cutout were we'd like no shadows to appear
3. Clear the stencil buffer, the Xbox version does this but not the PC version, this decouples the visible geometry's render states from affecting the shadow volumes (They should only depend on the Z-Buffer)
4. After drawing all the shadow volumes to the stencil buffer we draw our silhouette texture as a fullscreen quad with alpha testing enabled and z and color writes disabled, this discards any shadow that falls within our silhouette from the stencil buffer
5. Game takes over to draw shadows as normal

A big positive of these changes is we no longer have to go out of our way to detect James because he already has `STENCILOP_REPLACE` set. I'm hopeful other special shadow areas will now Just Work™, but if not, manual exclusion should be simpler due to the properly cleared stencil buffer.

I have adjusted the HeavensNightWindows, HotelHallwayWindow and WoodsideRoom208TV fixes to work with the new method and added in the hospital freezer interior, which just needs a Room ID check for safety.

This probably breaks some of our other fixes and needs testing, apologies in advance @Polymega :stuck_out_tongue:
Likely some D3D reference count issues too, sorry @elishacloud :stuck_out_tongue:

Test dll:
[xbox-shadows.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/3763726/xbox-shadows.zip)

An Xbox SH2 frame's draw order for reference:
Clear (RT/Stencil/Z) -> Map Geometry (Opaque) -> Dynamic Objects (Opaque) -> Dynamic objects (Transparent) -> Clear (Stencil) -> Shadow Volumes -> Shadows to backbuffer -> Map Geometry (Transparent) -> Fog overlay, lens flares, etc.

And a screenshot with color writes enabled to demonstrate it masking James and the window:
mask

### DIFF
--- a/Wrappers/d3d8/IDirect3DDevice8.cpp
+++ b/Wrappers/d3d8/IDirect3DDevice8.cpp
@@ -895,6 +895,11 @@ HRESULT m_IDirect3DDevice8::DrawPrimitive(D3DPRIMITIVETYPE PrimitiveType, UINT S
 			}
 		}
 
+		if (!silhouetteSurface)
+		{
+			return D3DERR_INVALIDCALL;
+		}
+
 		// Temporarily swap to new color buffer (keep original depth/stencil)
 		ProxyInterface->SetRenderTarget(silhouetteSurface, backbufferDepthStencil);
 		ProxyInterface->Clear(0, NULL, D3DCLEAR_TARGET, D3DCOLOR_ARGB(0, 0, 0, 0), 1.0f, 0);
@@ -925,7 +930,10 @@ HRESULT m_IDirect3DDevice8::DrawPrimitive(D3DPRIMITIVETYPE PrimitiveType, UINT S
 		// TODO: Should the soft shadow code do this too?
 		IDirect3DVertexBuffer8 *pStream0;
 		UINT stream0Stride = 0;
-		ProxyInterface->GetStreamSource(0, &pStream0, &stream0Stride);
+		if (SUCCEEDED(ProxyInterface->GetStreamSource(0, &pStream0, &stream0Stride)) && pStream0)
+		{
+			pStream0->Release();
+		}
 
 		// Discard all pixels but James' silhouette
 		DWORD stencilFunc = 0;
@@ -975,6 +983,11 @@ HRESULT m_IDirect3DDevice8::DrawPrimitiveUP(D3DPRIMITIVETYPE PrimitiveType, UINT
 			// We're done drawing shadow volumes, toggle flag off
 			shadowVolumeFlag = false;
 
+			if (!silhouetteTexture)
+			{
+				return D3DERR_INVALIDCALL;
+			}
+
 			// Bind our texture of James' silhouette
 			ProxyInterface->SetTexture(0, silhouetteTexture);
 
@@ -1004,7 +1017,10 @@ HRESULT m_IDirect3DDevice8::DrawPrimitiveUP(D3DPRIMITIVETYPE PrimitiveType, UINT
 			// TODO: Should the soft shadow code do this too?
 			IDirect3DVertexBuffer8 *pStream0;
 			UINT stream0Stride = 0;
-			ProxyInterface->GetStreamSource(0, &pStream0, &stream0Stride);
+			if (SUCCEEDED(ProxyInterface->GetStreamSource(0, &pStream0, &stream0Stride)) && pStream0)
+			{
+				pStream0->Release();
+			}
 
 			// Backup FVF, use pre-transformed vertices and texture coords
 			DWORD vshader = 0;
--- a/Wrappers/d3d8/IDirect3DDevice8.cpp
+++ b/Wrappers/d3d8/IDirect3DDevice8.cpp
@@ -400,7 +400,7 @@ HRESULT m_IDirect3DDevice8::SetRenderState(D3DRENDERSTATETYPE State, DWORD Value
 		{
 			if (SH2_SpecializedLight1 && *SH2_SpecializedLight1 != 0x01) // In a specialized lighting zone
 			{
-				if (SH2_RoomID && (*SH2_RoomID != 0x20 || *SH2_RoomID != 0x25 || *SH2_RoomID != 0x26)) // Exclude Blue Creek hallways/staircase
+				if (SH2_RoomID && *SH2_RoomID != 0x20 && *SH2_RoomID != 0x25 && *SH2_RoomID != 0x26) // Exclude Blue Creek hallways/staircase
 				{
 					Value = D3DSTENCILOP_ZERO;
 				}
--- a/Wrappers/d3d8/IDirect3DDevice8.cpp
+++ b/Wrappers/d3d8/IDirect3DDevice8.cpp
@@ -98,6 +98,16 @@ HRESULT m_IDirect3DDevice8::Reset(D3DPRESENT_PARAMETERS *pPresentationParameters
 		ReleaseInterface(&pOutTexture);
 	}
 
+	if (silhouetteSurface)
+	{
+		ReleaseInterface(&silhouetteSurface, 2);
+	}
+
+	if (silhouetteTexture)
+	{
+		ReleaseInterface(&silhouetteTexture);
+	}
+
 	// Update presentation parameters
 	UpdatePresentParameter(pPresentationParameters, nullptr, true);
 
@@ -386,7 +396,7 @@ HRESULT m_IDirect3DDevice8::SetRenderState(D3DRENDERSTATETYPE State, DWORD Value
 	// Restores self shadows
 	if (EnableSelfShadows && State == D3DRS_STENCILPASS && Value == D3DSTENCILOP_REPLACE)
 	{
-		if (/*SelfShadowTweaks &&*/ !SH2_SpecializedLight || *SH2_SpecializedLight != 0x01)
+		if (/*SelfShadowTweaks &&*/ SH2_SpecializedLight && *SH2_SpecializedLight != 0x01)
 		{
 			Value = D3DSTENCILOP_ZERO;
 		}
--- a/Wrappers/d3d8/IDirect3DDevice8.cpp
+++ b/Wrappers/d3d8/IDirect3DDevice8.cpp
@@ -883,7 +883,7 @@ HRESULT m_IDirect3DDevice8::DrawPrimitive(D3DPRIMITIVETYPE PrimitiveType, UINT S
 		}
 		if (SUCCEEDED(ProxyInterface->GetDepthStencilSurface(&backbufferDepthStencil)) && backbufferDepthStencil)
 		{
-			backbufferDepthStencil->Release(); // Is releasing this necessary when we set it again immediately?
+			backbufferDepthStencil->Release();
 		}
 
 		// Create texture for James' silhouette if it doesn't exist
@@ -1358,7 +1358,7 @@ HRESULT m_IDirect3DDevice8::Clear(DWORD Count, CONST D3DRECT *pRects, DWORD Flag
 	Logging::LogDebug() << __FUNCTION__;
 
 	// Change first Clear call to match Xbox version
-	if (Flags == (D3DCLEAR_TARGET | D3DCLEAR_STENCIL | D3DCLEAR_ZBUFFER) && Color == D3DCOLOR_ARGB(124, 0, 0, 0))
+	if (EnableXboxShadows() && Flags == (D3DCLEAR_TARGET | D3DCLEAR_STENCIL | D3DCLEAR_ZBUFFER) && Color == D3DCOLOR_ARGB(124, 0, 0, 0))
 	{
 		return ProxyInterface->Clear(Count, pRects, Flags, Color, Z, 0);
 	}
--- a/Wrappers/d3d8/IDirect3DDevice8.cpp
+++ b/Wrappers/d3d8/IDirect3DDevice8.cpp
@@ -382,19 +382,6 @@ HRESULT m_IDirect3DDevice8::SetRenderState(D3DRENDERSTATETYPE State, DWORD Value
 	{
 		Value = (Value * 15) / 16;
 	}
-	
-	// Restores self shadows
-	if (EnableSelfShadows && State == D3DRS_STENCILPASS && (Value == D3DSTENCILOP_ZERO || Value == D3DSTENCILOP_REPLACE))
-	{
-		if (SelfShadowTweaks &&	SH2_SpecializedLight && *SH2_SpecializedLight == 0x01)
-		{
-			Value = D3DSTENCILOP_REPLACE;
-		}
-		else
-		{
-			Value = D3DSTENCILOP_KEEP;
-		}
-	}
 
 	// For blur frame flicker fix
 	if (RemoveEffectsFlicker && State == D3DRS_TEXTUREFACTOR)
@@ -705,6 +692,26 @@ HRESULT m_IDirect3DDevice8::DrawIndexedPrimitive(THIS_ D3DPRIMITIVETYPE Type, UI
 {
 	Logging::LogDebug() << __FUNCTION__;
 
+	if (!shadowVolumeFlag)
+	{
+		// We're drawing opaque map geometry and dynamic objects
+
+		// Change default states to be more like Xbox version
+		ProxyInterface->SetRenderState(D3DRS_STENCILMASK, 0xFFFFFFFF);
+		ProxyInterface->SetRenderState(D3DRS_STENCILWRITEMASK, 0xFF);
+
+		DWORD stencilRef = 0;
+		if (SUCCEEDED(ProxyInterface->GetRenderState(D3DRS_STENCILREF, &stencilRef)) && stencilRef == 0)
+		{
+			// We're drawing just opaque map geometry
+
+			// Change default states to be more like Xbox version
+			ProxyInterface->SetRenderState(D3DRS_STENCILENABLE, TRUE);
+			ProxyInterface->SetRenderState(D3DRS_STENCILFUNC, D3DCMP_ALWAYS);
+			ProxyInterface->SetRenderState(D3DRS_STENCILPASS, D3DSTENCILOP_ZERO);
+		}
+	}
+
 	// Exclude Woodside Room 208 TV static geometry from receiving shadows
 	if (WoodsideRoom208TV && SH2_RoomID && *SH2_RoomID == 0x18 && Type == D3DPT_TRIANGLESTRIP && MinVertexIndex == 0 && NumVertices == 4 && startIndex == 0 && primCount == 2)
 	{
@@ -723,7 +730,7 @@ HRESULT m_IDirect3DDevice8::DrawIndexedPrimitive(THIS_ D3DPRIMITIVETYPE Type, UI
 			DWORD stencilPass = 0;
 			ProxyInterface->GetRenderState(D3DRS_STENCILPASS, &stencilPass);
 
-			ProxyInterface->SetRenderState(D3DRS_STENCILPASS, D3DSTENCILOP_ZERO);
+			ProxyInterface->SetRenderState(D3DRS_STENCILPASS, D3DSTENCILOP_REPLACE);
 
 			hr = ProxyInterface->DrawIndexedPrimitive(Type, MinVertexIndex, NumVertices, startIndex, primCount);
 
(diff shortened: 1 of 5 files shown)


Polymega:
giphy

You wonderful person, you!

Polymega:
This is so rad, FrozenFish! I have to say: Excluding the shadows in the fridge is another one of those small but quite impactful changes. This is such cool work!

> This probably breaks some of our other fixes and needs testing, apologies in advance @Polymega 

Not a single worry here, my friend! Happy to go back and test things. Off the top of your head: Are there any particular fixes I should pay extra attention to when going back for testing? 

> A big positive of these changes is we no longer have to go out of our way to detect James because he already has STENCILOP_REPLACE set. I'm hopeful other special shadow areas will now Just Work™, but if not, manual exclusion should be simpler due to the properly cleared stencil buffer.

_So sexy..._ In your proof-of-concept build should James have restored self shadows when he's not in specialized lighting zones? Or are his self shadows fully disabled for this test? 

-----------------

@elishacloud an early note: If you start the game then launch another piece of software that prompts the UAC it'll freeze/soft lock the game. We ran into this issue before, so it'll need to be addressed again for the new code. 

> and added in the hospital freezer interior, which just needs a Room ID check for safety.

The Room ID for the Hospital interior fridge shadow exclusion is `0x53`.

elishacloud:
Nice work @FrozenFish24!

> Likely some D3D reference count issues too, sorry.

No problem.  I can fix these issues.

> If you start the game then launch another piece of software that prompts the UAC it'll freeze/soft lock the game. We ran into this issue before, so it'll need to be addressed again for the new code.

Yes, some code needs to be added to the `Reset()` function to fix this issue.  I'll do a bit more work on this code once I check it in.

FrozenFish24:
@Polymega 
> Not a single worry here, my friend! Happy to go back and test things. Off the top of your head: Are there any particular fixes I should pay extra attention to when going back for testing?

Anything involving shadows or detecting geometry. Looks like I broke the top down shadow fix for instance. Anything that breaks in one place should break everywhere so it shouldn't be necessary to check more than one representative area.

> _So sexy..._ In your proof-of-concept build should James have restored self shadows when he's not in specialized lighting zones? Or are his self shadows fully disabled for this test?

I was hoping he would but looks like it was wishful thinking on my part that they left the Xbox render states intact. I've fixed this in the latest commit, he should have his self-shadows back in the proper spots.

> The Room ID for the Hospital interior fridge shadow exclusion is `0x53`.

Thanks! I've added a check for the room ID and a temporary ini option like the other fixes. Add `HospitalRefrigerator = 1` to stop it receiving shadows.

@elishacloud 
>No problem. I can fix these issues.

Awesome! Glad one of us knows what we're doing :P

Test dll with changes:
[xbox-shadows-v2.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/3768046/xbox-shadows-v2.zip)

Polymega:
@FrozenFish24 
Your revised code here is incredibly solid! In the main campaign, self shadows behave virtually 1:1 to the PS2 version. Well, _well_ done!! I ran through both campaigns and here are my notes from it: 

-------------

The "solid" top-down shadow code isn't working for either of the Eddie boss rooms (`0x8F, 0x90`).
- It is working fine for the Observation Deck (`0x02`) and Angela Apartment Room (`0x24`), along with the Leave ending cutscene (`Cutscene 0x5A`), still.

-------------

Fully **disable** James' self shadow while in Rooms `0x9E, 0x9A, 0xB3, 0xAF`.
- This is no fault of the code, but buggy lighting behavior inherent to these rooms in relation to James' self shadow. These shadow bugs either affect all versions of the game or just the PC version.
   - I plan to do one more sweep later on to ensure there's no other bugged rooms like this. It should be easy to add any extra room IDs, once the initial code for this is added in.

-------------

Add an additional window to geometry that shouldn't receive any shadows on it. This is for **rr55.map** (room ID `0x94`): 

sh2pc 2019-10-24 16-34-33

-------------

For the Born From a Wish side campaign: Maria should have self shadows within many of the rooms. This new code disables her self shadows except for during cutscenes. 

Potential psuedo-code: 
- [If Born From a Wish campaign is active](https://github.com/elishacloud/Silent-Hill-2-Enhancements/blob/master/Wrappers/d3d8/IDirect3DDevice8.cpp#L1500) **AND NOT** in Room `0x20, 0x25, 0x26`:
   - Disable the new self shadow code (to restore her self shadow for her side campaign).

We add the "AND NOT in Room" part because, in these rooms, her self shadow should be disabled, which it is with the new, current self shadow code.

-------------

...and that's all I got!

FrozenFish24:
@Polymega 
I've added an exception for the window in rr55

>For the Born From a Wish side campaign: Maria should have self shadows within many of the rooms. This new code disables her self shadows except for during cutscenes.

Not sure if I'm understanding you correctly here, I played the start of BFaW and Maria's self shadows work correctly for me in Heaven's Night and the Baldwin House.

I'm a bit busy the next few days but I'll look into the other issues as soon as possible.

Polymega:
> Not sure if I'm understanding you correctly here, I played the start of BFaW and Maria's self shadows work correctly for me in Heaven's Night and the Baldwin House.

My apologies. I must've disabled self shadows to test something near the end, forgot I did that, and then played the BFaW campaign. You're right; she has self shadows in her campaign. Sorry for the confusion there and thanks for catching that.

I've edited my earlier post to reflect this. So the info below would be an icing-on-top request then: 

When Team Silent made the BFaW campaign they experimented with more advanced shadow behaviors. In a few rooms only, her self shadow will smoothly fade in-and-out as she moves around in these specialized lighting zones. ([Similar to how Haunted Ground's shadows behave.](https://youtu.be/Ektkbejc-sE?t=971))

This self shadow fading behavior doesn't seem to be preserved for the PC build. And while it still looks fine overall (even without the fade), there are a few particular rooms where her self shadow behavior _does_ look rough without that fade bevahior. 

If possible, can self shadows be disabled for the following rooms? 

- [If Born From a Wish campaign is active](https://github.com/elishacloud/Silent-Hill-2-Enhancements/blob/master/Wrappers/d3d8/IDirect3DDevice8.cpp#L1500) **AND** in Room `0x20, 0x25, 0x26`:
   - Disable self shadows - For my tests in these rooms I set `EnableSelfShadows = 0` in the ini.

This is for both Blue Creek hallways + Blue Creek stairway. The removal of self shadows for these rooms would likely go unnoticed by most/all.

Polymega:
> Fully disable James' self shadow while in Rooms ...

Ignore this original request as well. It's actually a larger problem that needs a universal fix, as it happens in all special light zones: If you walk backwards into a specialized light zone James will have a self shadow for just a momentary flicker. 

You can straddle the edge of a special light zone to trigger it repeatedly by leaning forwards. This only happens when the flashlight is off: 
https://youtu.be/sAbZl1gFqnE

I cross-checked with PS2 and Xbox and this doesn't happen there.

I looked at `00932B50 (byte)` (what we're using to determine when he enters/leaves a special light zone) and the value isn't "flickering" along a light zone's edge, as far as I can tell? Maybe it is and Cheat Engine isn't displaying it fast enough for me? Maybe we need to use another address for this that maybe changes its value a little sooner (if such one exists)?

Not sure about this one at the moment...

elishacloud:
> If you start the game then launch another piece of software that prompts the UAC it'll freeze/soft lock the game.

@FrozenFish24, adding the following lines into the `Reset()` function will fix this issue:

```c++
if (silhouetteSurface)
{
	ReleaseInterface(&silhouetteSurface, 2);
}

if (silhouetteTexture)
{
	ReleaseInterface(&silhouetteTexture);
}
```

Wrappers/d3d8/IDirect3DDevice8.cpp
```diff
@@ -382,17 +382,13 @@ HRESULT m_IDirect3DDevice8::SetRenderState(D3DRENDERSTATETYPE State, DWORD Value
...
 	// Restores self shadows
-	if (EnableSelfShadows && State == D3DRS_STENCILPASS && (Value == D3DSTENCILOP_ZERO || Value == D3DSTENCILOP_REPLACE))
+	if (EnableSelfShadows && State == D3DRS_STENCILPASS && Value == D3DSTENCILOP_REPLACE)
 	{
-		if (SelfShadowTweaks &&	SH2_SpecializedLight && *SH2_SpecializedLight == 0x01)
-		{
-			Value = D3DSTENCILOP_REPLACE;
-		}
-		else
+		if (/*SelfShadowTweaks &&*/ !SH2_SpecializedLight || *SH2_SpecializedLight != 0x01)
```
elishacloud: This line needs to be changed.  I it should look like the following.  We need to check if the pointer is good before using it.

```c++
if (SH2_SpecializedLight && *SH2_SpecializedLight != 0x01)
```

Also, you are removing the `SelfShadowTweaks` option but not adding it back anywhere.  Maybe you should remove this option from the settings files and replace it with a new temp option to allow enabling and disabling self shadows in specialized light.

FrozenFish24: If `SH2_SpecializedLight` is null the if statement will short circuit and it won't dereference the pointer.

With my code if `GetSpecializedLightPointer()` fails and returns `nullptr` James will always self-shadow, with your code if it fails James will never self-shadow. These are edge cases though and yours is clearer so I'm happy to switch.

>Also, you are removing the SelfShadowTweaks option but not adding it back anywhere.

Good call, I'll add an option in my next commit.

FrozenFish24:
@Polymega
I believe the cause of the shadow flicker was because while the value at `00932B50` WAS being set to `1` on the first frame of entering a special lighting zone, it was being set after the call to `SetRenderState(D3DRS_STENCILPASS, D3DSTENCILOP_REPLACE)` so James had `ZERO` set for a single frame.

I've located another address that changes earlier on in the game loop, so this is no longer an issue. Also looking at the assembly these addresses should probably be DWORDS so I changed the return type of `GetSpecializedLightPointer()`.

I've also added @elishacloud's code to `Reset()`

>If Born From a Wish campaign is active AND in Room 0x20, 0x25, 0x26:
Disable self shadows - For my tests in these rooms I set EnableSelfShadows = 0 in the ini.

Would you happen to have saves for these areas? I want check my conditional logic is correct before committing it.

Will get the ini option and top down shadow fix underway soon.

test dll:
[xbox-shadows-v3.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/3783286/xbox-shadows-v3.zip)

Polymega:
> I've located another address that changes earlier on in the game loop, so this is no longer an issue.

Great! I'll give this another test soon and report back. Thank you as always, FrozenFish. :)

> Would you happen to have saves for these areas? I want check my conditional logic is correct before committing it.

Yes, sir: [Folder 01.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/3784180/Folder.01.zip) (The save near the bottom.) 

Here is what you'd be testing to disable for the Blue Creek hallways/staircase in Maria's campaign: https://youtu.be/UUMxWJlL2iA

Also, while you're coding this part, may I ask you to also disable James' self shadow while in Hotel Room 202-204 (Room ID `0x9E`). This is an inherent self shadow bug that affects both the PC and PS2 versions (and probably the Xbox version, too). I believe this is the only bugged-self-shadow-with-flashlight-on room in the PC version.

sh2pc 2019-10-29 10-56-39

FrozenFish24:
>One thing we could try is utilizing both of these addresses

Smart thinking. Looks like this is a good solution. The two areas in your earlier YT videos no longer flicker, and I found a zone that broke with my address in the hotel lobby that now works again when we check both. Give it a try:

[xbox-shadows-v4.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/3788754/xbox-shadows-v4.zip)

Polymega:
Nice. I anticipate this will fully fix self shadow behavior so I'm going to do another complete, slow play test. Please give me a day or three with this and I'll report back.

Polymega:
This actually went much quicker than I anticipated simply due to how well it all worked. 

Soft shadow behavior within the main campaign is fantastic now! Great catch with that new address to remove the flicker when James enters/leaves light zones. :D

Here's my one-and-only new note:

- For the BFaW side campaign: In this test build (v4), it disables Maria's self shadow in a variety of areas where she should keep it. In v2 of your test build, her self shadows were correct during her campaign. I'm assuming the introduction of the new address to check whenever James is in a special light zone might be the culprit? If so, then disable these address checks for light zones while BFaW is active?

...and a recap of final adjustments:

- Restore solid top-down shadow for Eddie boss rooms (`0x8F, 0x90`)
- Disable James' self shadow while in Hotel Rm 202-204 (`0x9E`)
- Disable self shadows for BFaW while in Blue Creek hallways/staircase (`0x20, 0x25, 0x26`)
- Ensure game does not freeze/soft lock if a UAC prompt happens after game launch with new code (game still froze for me when this happened)

elishacloud:
> Ensure game does not freeze/soft lock if a UAC prompt happens after game launch with new code (game still froze for me when this happened)

@FrozenFish24, if you have an issue with this one, I can fix this after you check-in the other fixes.  I believe this has to do with the `Reset()` function failing.  At least that's what it was last time.

FrozenFish24:
I've removed the check on the second address while playing BFaW. Will this introduce flickering? Or did BFaW never flicker to begin with?

I added exclusions for `0x9E` and BFaW `0x20`, `0x25`, `0x26`

I've also added a new ini option to completely enable/disable all my new code `XboxShadows = 1` and disabled it in all top down shadow rooms/cutscenes. Interestingly this still doesn't fix the Eddie boss rooms. I think this might tie into the `Reset()` issues with something not being released/restored correctly.

Am I right in assuming we'll collapse down the following options into one once everything works?

`EnableSelfShadows`
`HeavensNightWindows`
`HospitalRefrigerator`
`HotelHallwayWindow`
`HotelStoreroomWindow`
`LabyrinthValveTurn`
`SelfShadowTweaks`
`ShadowIntensity`
`WoodsideRoom208TV`
`XboxShadows`

And these will be kept separate:
`EnableSoftShadows`
`TopDownShadow`

test dll:
[xbox-shadows-v5.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/3794075/xbox-shadows-v5.zip)

elishacloud:
> Am I right in assuming we'll collapse down the following options into one once everything works?
> 
> `EnableSelfShadows`
> `HeavensNightWindows`
> `HospitalRefrigerator`
> `HotelHallwayWindow`
> `HotelStoreroomWindow`
> `LabyrinthValveTurn`
> `SelfShadowTweaks`
> `ShadowIntensity`
> `WoodsideRoom208TV`
> `XboxShadows`

Correct.  We plan on merging these into a single option later.  I just added them as separate options to make it easier for @Polymega to test.

> And these will be kept separate:
> `EnableSoftShadows`
> `TopDownShadow`

These will probably be merged as well.  We will likely end up with just one option: `EnableSoftShadows`

> Interestingly this still doesn't fix the Eddie boss rooms. I think this might tie into the `Reset()` issues with something not being released/restored correctly.

I don't see how this could be the `Reset()` issue.  However, I can look into this and the UAC issue and see what is going on.

FrozenFish24:
>I don't see how this could be the Reset() issue. However, I can look into this and the UAC issue and see what is going on.

Thanks for looking into the UAC stuff @elishacloud. :+1: 

I thought the Eddie shadows might be an issue with reference count causing a buffer to be freed when it shouldn't or something (I don't really know what I'm talking about :P). Turned out I was still intercepting `Clear()` even when `XboxShadows = 0` or in a top down shadow room. Disabling that along with everything else fixed the Eddie rooms.

Polymega:
We're right at the finish line now! 

:ballot_box_with_check: Restore solid top-down shadow for Eddie boss rooms (0x8F, 0x90)
:ballot_box_with_check: Disable James' self shadow while in Hotel Rm 202-204 (0x9E)
:ballot_box_with_check: Disable self shadows for BFaW while in Blue Creek hallways/staircase (0x20, 0x25, 0x26)

> I've removed the check on the second address while playing BFaW. Will this introduce flickering? Or did BFaW never flicker to begin with?

I just quickly went through the side campaign again and self shadow behavior seems fine for BFaW still (in regards to flickering). I think its fine because she was never meant to have her self shadows disabled in special light zones--like what we must do for James--so flickering is a non-concern for her campaign. 

Although, TBH, shadow behavior in general for BFaW is kind of whack... Team Silent used many, _many_ different light sources within any given room (mostly within the mansion) and shadow behavior can get a little wild. 

For BFaW: My main goal here is to get shadow behavior to match the PS2 version as close as possible and call it a day, due to how wild these light/shadow sources inherently are for this campaign. The only exception is places like the Blue Creek hallways/staircase, since our version won't smoothly fade out the self shadows in these areas, so the self shadow behavior here is a little too jarring.

-----------------

The only thing is that Maria's self shadow is disabled in various spots within her BFaW campaign, still. I assumed it might have been the introduction of the additional light zone address check, but maybe that's not the reason why? It works in v2, as a point of reference. An easy test is to do the following: 

- Load the save point "**Maria: Mansion Hallway**"
- Head to the "Study" room (two rooms over)
- As soon as you enter this room you should see a self shadow immediately on Maria. Move her around or make her squat to see it in motion to tell easier. 

Here's a screen cap using v2 of the test build at this location: 

sh2pc-2019-10-31-14-47-11

FrozenFish24:
Whoops, my bad. I had the two pointers mixed up. This will fix it:
[xbox-shadows-v7.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/3795439/xbox-shadows-v7.zip)

Thanks for suffering through all this testing and re-testing @Polymega :smile:

Polymega:
That's the ticket! She also has self shadows in the Blue Creek areas again but let's not worry about that; it's too small of a worry in the grand scheme of things here to spend much more time on.

> Thanks for suffering through all this testing and re-testing

No need to thank me; I'm the one who should (and always will!) thank _and_ apologize for making you guys tweak everything to the moon and back.

So, thanks for bearing with me and all my notes. That goes out to all of you wonderful teammates on this project. :heart:

Whenever you're ready and feel good about it, I think it's ready to be committed! You have now restored soft shadows and self shadows. FrozenFish...
### You're awesome!

FrozenFish24:
>She also has self shadows in the Blue Creek areas again

Another derp on my part. Had ORs instead of ANDs:
[xbox-shadows-v8.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/3795604/xbox-shadows-v8.zip)

>You're awesome!

Right back at you. Having someone who knows the game inside and out and willing to put in hours cataloging, testing and retesting every little change has made doing these fixes so much easier! Kudos! :tada:

Polymega:
Stick a fork in it, my friend, as it's done!!

elishacloud:
Great work @FrozenFish24!  I just have a few changes coming for you.

Wrappers/d3d8/IDirect3DDevice8.cpp
```diff
@@ -805,19 +866,198 @@ HRESULT m_IDirect3DDevice8::DrawPrimitive(D3DPRIMITIVETYPE PrimitiveType, UINT S
...
+		if (!silhouetteTexture)
+		{
+			if (SUCCEEDED(ProxyInterface->CreateTexture(BufferWidth, BufferHeight, 1, D3DUSAGE_RENDERTARGET, D3DFMT_A8R8G8B8, D3DPOOL_DEFAULT, &silhouetteTexture)) && silhouetteTexture)
+			{
+				silhouetteTexture->GetSurfaceLevel(0, &silhouetteSurface);
+			}
+		}
+
+		// Temporarily swap to new color buffer (keep original depth/stencil)
+		ProxyInterface->SetRenderTarget(silhouetteSurface, backbufferDepthStencil);
```
elishacloud: `silhouetteSurface` is used for the first time here, but it is never validated before it is used.  If `GetSurfaceLevel` fails for whatever reason the game will crash here.

Wrappers/d3d8/IDirect3DDevice8.cpp
```diff
@@ -805,19 +866,198 @@ HRESULT m_IDirect3DDevice8::DrawPrimitive(D3DPRIMITIVETYPE PrimitiveType, UINT S
...
+		ProxyInterface->GetRenderState(D3DRS_ALPHABLENDENABLE, &alphaBlend);
+		ProxyInterface->GetRenderState(D3DRS_ALPHATESTENABLE, &alphaTest);
+		ProxyInterface->SetRenderState(D3DRS_ALPHABLENDENABLE, FALSE);
+		ProxyInterface->SetRenderState(D3DRS_ALPHATESTENABLE, FALSE);
+
+		// DrawPrimitiveUP trashes vertex buffer stream 0, back it up
+		// TODO: Should the soft shadow code do this too?
+		IDirect3DVertexBuffer8 *pStream0;
+		UINT stream0Stride = 0;
+		ProxyInterface->GetStreamSource(0, &pStream0, &stream0Stride);
```
elishacloud: This line will cause a reference counter issue.  It should be changed to:

```c++
if (SUCCEEDED(ProxyInterface->GetStreamSource(0, &pStream0, &stream0Stride)) && pStream0)
{
	pStream0->Release();
}
```

Wrappers/d3d8/IDirect3DDevice8.cpp
```diff
@@ -805,19 +866,198 @@ HRESULT m_IDirect3DDevice8::DrawPrimitive(D3DPRIMITIVETYPE PrimitiveType, UINT S
...
+			ProxyInterface->GetRenderState(D3DRS_ALPHABLENDENABLE, &alphaBlend);
+			ProxyInterface->GetRenderState(D3DRS_ALPHATESTENABLE, &alphaTest);
+			ProxyInterface->SetRenderState(D3DRS_ALPHABLENDENABLE, FALSE);
+			ProxyInterface->SetRenderState(D3DRS_ALPHATESTENABLE, TRUE);
+
+			// DrawPrimitiveUP trashes vertex buffer stream 0, back it up
+			// TODO: Should the soft shadow code do this too?
+			IDirect3DVertexBuffer8 *pStream0;
+			UINT stream0Stride = 0;
+			ProxyInterface->GetStreamSource(0, &pStream0, &stream0Stride);
```
elishacloud: This line will cause a reference counter issue.  It should be changed to:

```c++
if (SUCCEEDED(ProxyInterface->GetStreamSource(0, &pStream0, &stream0Stride)) && pStream0)
{
	pStream0->Release();
}
```

FrozenFish24:
Thanks @elishacloud! Added your code and UAC no longer soft locks the game :+1:

elishacloud:
This update was much more involved then I first thought.  Really good work @FrozenFish24!

PR #207, opened 2020-01-27
Right analog stick support for search camera

CookiePLMonster:
This PR changes the way search camera works with gamepads, restoring right analog stick functionality like on PS2.

Most of the code related to right analog stick was still left in the executable, likely a leftover from PS2. Changes required are:
1) Right stick is now polled from DirectInput.
2) Search camera is now bound to right stick instead of left stick. Thankfully, keyboard controls have been nicely separated so this change doesn't impact KB/M users.

Additionally, "Search camera" toggle now allows the player to walk around when using a gamepad. Keyboard functionality is kept untouched.

`RestoreSearchCamMovement` INI option has been added with multiple options, allowing the player to tailor controls to their specific gamepad - those options mirror the choices XInput Plus gives users when mapping DInput gamepads to XInput:
- 1 - right stick uses X/Y Rotation value (XInput gamepads map to this)
- 2 - right stick uses Z axis/Z Rotation
- 3 - right stick uses Z rotation/Z axis

Who knows, perhaps this very lack of consistency was the reason this feature was cut from PC versions...?

Additionally, a Southpaw option has been added, which swaps all functions of left and right analog sticks. It's a fully polished option present in the game, but it went fully unused, even on PS2.

Comments are suggestions are welcome.

### DIFF
--- a/Common/Settings.h
+++ b/Common/Settings.h
@@ -67,6 +67,7 @@
 	visit(Room312ShadowFix, true) \
 	visit(RowboatAnimationFix, true) \
 	visit(SetBlackPillarBoxes, true) \
+	visit(Southpaw, false) \
 	visit(SteamCrashFix, true) \
 	visit(UnlockJapLang, false) \
 	visit(UseCustomExeStr, true) \
--- a/Common/Settings.h
+++ b/Common/Settings.h
@@ -90,6 +90,7 @@
 	visit(ResY, 0) \
 	visit(FPSLimit, 30) \
 	visit(IncreaseNoiseEffectRes, 512) \
+	visit(RestoreSearchCamMovement, 1) \
 	visit(SingleCoreAffinity, 1) \
 	visit(SingleCoreAffinityTimer, 5000) \
 	visit(SmallFontWidth, 14) \
--- a/Common/Settings.ini
+++ b/Common/Settings.ini
@@ -54,6 +54,9 @@ DPadMovementFix = 1
 ; Allows the same button to be assigned to multiple actions on controllers. (0|1)
 GamepadControlsFix = 1
 
+; Sets right joystick mode for search camera movement on controllers. 1 = XInput (most common), 2 = DirectInput, 3 = DirectInput (less common). (0|1|2|3)
+RestoreSearchCamMovement = 1
+
 ; Restores vibration for controllers. (0|1)
 RestoreVibration = 1
 
--- a/Common/Settings.ini
+++ b/Common/Settings.ini
@@ -60,6 +60,9 @@ RestoreSearchCamMovement = 1
 ; Restores vibration for controllers. (0|1)
 RestoreVibration = 1
 
+; Swaps left and right joystick functions. Useful for left handed players. (0|1)
+Southpaw = 0
+
 ; Pad (controller) ID to vibrate. Default pad ID = 0. (0|*)
 PadNumber = 0
 
--- a/Patches/ControllerTweaks.cpp
+++ b/Patches/ControllerTweaks.cpp
@@ -20,6 +20,7 @@
 #include "Common\Utils.h"
 #include "Common\Settings.h"
 #include "Logging\Logging.h"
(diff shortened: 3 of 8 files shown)


elishacloud:
Thanks for your work on this, @CookiePLMonster!  I will review the code and let you know my comments soon.

> Patterns used here are really painful to work with (no wildcards...) so it takes me more time than I'd like.

I just recently included ThirteenAG's [Hooking.Patterns](https://github.com/ThirteenAG/Hooking.Patterns) so that I can compile the [WidescreenFixesPack](https://github.com/ThirteenAG/WidescreenFixesPack).  Hooking.Patterns supports wildcards.  I don't currently use it anywhere except the WidescreenFixesPack, but you are welcome to use it if it would make things easier for you.

To use it you just add [`#include "External\Hooking.Patterns\Hooking.Patterns.h"`](https://github.com/elishacloud/Silent-Hill-2-Enhancements/blob/fc3694362a7bac026691bc8048b3894633d86d21/Include/GTA/WSF.h#L27).  You can see an example of it in use [here](https://github.com/ThirteenAG/WidescreenFixesPack/blob/6ee7da872aba5586737199bc692e897a59bab66c/source/SilentHill2.WidescreenFix/dllmain.cpp#L90).

CookiePLMonster:
> I just recently included ThirteenAG's [Hooking.Patterns](https://github.com/ThirteenAG/Hooking.Patterns) so that I can compile the [WidescreenFixesPack](https://github.com/ThirteenAG/WidescreenFixesPack). Hooking.Patterns supports wildcards. I don't currently use it anywhere except the WidescreenFixesPack, but you are welcome to use it if it would make things easier for you.

Yeah they absolutely would. That's also what I initially suggested, given that my patterns library (https://github.com/CookiePLMonster/ModUtils/blob/master/Patterns.h) and his patterns library are both forks of the same code + ThirteenAG rebased his work on top of mine at some point - so I am perfectly accustomed with the way they should be used and they would be most convenient for me. The biggest win is that the format these patterns use is undestood by IDA so you can just copy those strings and search for them directly.

CookiePLMonster:
I added support for multiple executables. I still need to fix strafing before I finalize this PR.

elishacloud:
> I added support for multiple executables. I still need to fix strafing before I finalize this PR.

Great.  Everything is looking good so far.

> I strongly encourage having this factorized into a function (like in my ModUtils lib - https://github.com/CookiePLMonster/ModUtils/blob/master/MemoryMgr.h#L117) because right now it leaves way too much room for error - I did not get the maths right the first time now either.

Any suggestions here are appreciated.  Can you explain this in a bit more detail?  How does your `ModUtils` lib help reduce errors?

CookiePLMonster:
> Any suggestions here are appreciated. Can you explain this in a bit more detail? How does your `ModUtils` lib help reduce errors?

Take a look eg. at https://github.com/elishacloud/Silent-Hill-2-Enhancements/pull/207/files#diff-6ec9705b12a871539efb545de53d8f42R146 - you need to maintain proper calculations and get proper types (for example if `jmpAddress` was a DWORD* then this code will still compile but will yield wrong results!). Having it factorized into a `ReadCall`-like function would remove any risk like this.

CookiePLMonster:
I have submitted a fix for strafing and also added a Southpaw option.

Currently strafing is enabled for both gamepad and keyboard, as I couldn't figure out how to check for gamepad explicitly in this context. It's not a big issue though.

**This PR is now finished.**

elishacloud:
> Take a look eg. at https://github.com/elishacloud/Silent-Hill-2-Enhancements/pull/207/files#diff-6ec9705b12a871539efb545de53d8f42R146 - you need to maintain proper calculations and get proper types (for example if `jmpAddress` was a DWORD* then this code will still compile but will yield wrong results!). Having it factorized into a `ReadCall`-like function would remove any risk like this.

I see.  I will look into adding something like that.  Thanks!

> This PR is now finished.

Nice work!  Looks good.

Issue #275, opened 2020-05-04
Main Menu/Options Menu Adjustments


Polymega:
Hi @belek666,

-------------------

### Options menu adjustment

**Update:** Aero_ found a solution to this request: https://github.com/elishacloud/Silent-Hill-2-Enhancements/issues/199#issuecomment-623694764

~~In the options menu for the game, for "Weapon Control" and "View Control", the two selections are 'Normal' and 'Switch'.~~

~~Inside `option_msg_*.mes` there is an unused string that says "Press".~~

![image](https://user-images.githubusercontent.com/27395260/81010351-25a8ae80-8e24-11ea-842a-3c9efcbc3575.png)

~~Can the selections for "Weapon Control" and "View Control" be changed to 'Press' and 'Switch'?
"Press" should have been the correct selection this whole time.~~

Thank you to Badore90 for finding the unused "Press" string for use here.

--------------------------------

### Main menu adjustment

This is in regards to the `MainMenuTitlePerLang` feature.

If you enable the Japanese language selection (`UnlockJapLang = 1`) and select the Japanese language in the options menu, the main menu will then look like this:

![image](https://user-images.githubusercontent.com/27395260/81009728-25f47a00-8e23-11ea-93cb-b1e3a98e3d57.png)

... unless you have a `start01j.tex` file present in `SILENT HILL 2\sh2\pic\etc\`. 

You created a failsafe check that, if the correct `start01*.tex` is **not** present, it will default to using `start01.tex` instead. I believe this safety check might have been excluded for the Japanese language selection. [Possibly in this portion of code?](https://github.com/elishacloud/Silent-Hill-2-Enhancements/blob/master/Patches/MainMenu.cpp#L65-L74)

belek666:
Hello @Polymega 
The problem does not exist when not using hi res menu textures. By looking at texture patching code in `TexPatch.cpp` I can say that `start01.tex` resolution is read only once at start and texture scale values are calculated at this time. Solution would be to update those variables every Main Menu call. The `ScaleStart01Texture` function (with small modification - writing jumps to ASM code only once) could be called in function `UpdateTitlePath` in `MainMenu.cpp`.

Polymega:
Thank you, Belek. Unfortunately, I still have the error.

[...]

PR #277, opened 2020-05-10
Update start01.tex scale values every Main Menu call

belek666:
Fix related to issue https://github.com/elishacloud/Silent-Hill-2-Enhancements/issues/275#issue-612144058


Polymega:
Thank you, Belek. Another added benefit to this, that I just realized, is that it will help with fan translation packs that predate this project. Meaning, old translation packs that would use the original, low res `start01.tex` file will still display correctly.

Issue #73, opened 2018-12-14
Restoring specularity to objects throughout the game


Polymega:
This may be a more involved fix so making its own post. 

Throughout the game, certain meshes should always have a specularity/"reflection" to them. These include James' eyes; flashlight body; flashlight lens; all other character eyes; Maria's skirt, necklace, hip rope/pendants; etc. 

Yet, throughout the game these meshes will often not have any specularity to them. It's only when you go in specific rooms or when specific cutscenes initiate that these types of objects get their shine back. 

Here are some image examples of where they shine: https://github.com/elishacloud/Silent-Hill-2-Enhancements/issues/66#issuecomment-443391958

Looking into this I've noticed shader consistencies between when these things do and don't shine. This was tested by capturing a handful of scenes using the Intel Frame Analyzer where certain mesh objects do shine correctly and where they don't. I haven't done any extensive research with this yet but captured a few test areas for the time being.  

1. I don't believe this is a true specular material but instead something that mimics the effect. When the "specularity" works, the mesh has three textures: 1) The actual texture file, 2) a simple white gradient circle (that very likely has some sort of alpha channel with it) which simulates the specular shine, and 3) a cube map for which the gradient circle moves around in relation to the camera's position upon the mesh. 

![image](https://user-images.githubusercontent.com/27395260/49977182-53c9dd80-ff13-11e8-8e12-923510079cbc.png)

2. The gradient circle is always the second texture loaded in, and the cube map is always the last (third) texture to load in. 

3. Every object that does shine may have different vertex shader codes, but there are portions within each of their vertex shader codes that share identical lines to others that also render the specularity right.

4. The majority of objects that do shine have the same, exact pixel shader code. 

5. Objects that should shine but don't do not have any pixel shader. They do have a vertex shader and, as mentioned above, there are lines within it that are indentical to those in vertex shaders where the shine works. 

6. Objects that should shine but don't do not have a cube map; just the main texture file and the white gradient: 

![image](https://user-images.githubusercontent.com/27395260/49978402-5a0e8880-ff18-11e8-8e6d-cadcb4566046.png)

------------------------

For the majority of meshes that do correctly shine, they all share the following pixel shader:

**Edit:** Removed _a bunch_ of vertex shader code that really didn't mean anything.

-------------------------------------

All just observational talk at the moment but perhaps worth sharing the segments of similarites in shaders between different meshes that currently do shine correctly vs. those that don't.

[...]

PR #381, opened 2021-02-06
Specular Fix

FrozenFish24:
Belated pull request for the [specularity fixes](https://github.com/elishacloud/Silent-Hill-2-Enhancements/issues/73) we mostly finished up in December.

spec

Restores specularity to characters and objects throughout the game.
Also modifies the Woodside Room 208 TV fix to use the new `GetModelID()` so it plays nice with `IncreaseNoiseEffectRes`

Still fairly rough, so marking as Draft.
**Please don't hesitate to post feedback and nitpicks or ask for further clarification.**

# Explanation:
Objects/characters will only exhibit specularity when lit by a `D3DLIGHT_DIRECTIONAL` light source. If no directional lights are found for the current object/character we add our own.

For greater context check out the partial decompilations of the related functions [here](https://github.com/FrozenFish24/Silent-Hill-2-Enhancements/blob/d875c8f35e10958a433398d327ef92aac1ce2eb2/Patches/actorDrawOpaque_501540.cpp), [here](https://github.com/FrozenFish24/Silent-Hill-2-Enhancements/blob/d875c8f35e10958a433398d327ef92aac1ce2eb2/Patches/setupActorDrawOpaque_4FE850.cpp), [here](https://github.com/FrozenFish24/Silent-Hill-2-Enhancements/blob/d875c8f35e10958a433398d327ef92aac1ce2eb2/Patches/setupActorDrawTransparent_502220.cpp), [here](https://github.com/FrozenFish24/Silent-Hill-2-Enhancements/blob/d875c8f35e10958a433398d327ef92aac1ce2eb2/Patches/sub_501F40.cpp) and [here](https://github.com/FrozenFish24/Silent-Hill-2-Enhancements/blob/d875c8f35e10958a433398d327ef92aac1ce2eb2/Patches/sub_50EA80.cpp).

The patch is performed in 5 parts in which we hook one of the game's original functions with some extra code of our own (function names are arbitrary):

## Part 1: HookActorDrawTop
Hooks a call to `void ActorDrawTop(ModelOffsetTable* pOffsetTable, void* arg2)`

The material count and pointer to the material array are recorded here, so we can traverse it later (for special casing opacities).

## Part 2: HookGetLightSourceCount
Hooks a call to `int GetLightSourceCount()`

Searches the LightSource array for a `D3DLIGHT_DIRECTIONAL`, if none is present return the real lightsource count + 1 and set flag so we know to return our fake later. We also check for `D3DLIGHT_SPOT` here which lets us know we're in a special light zone.

## Part 3: HookGetLightSourceAt
Hooks a call to `LightSourceStruct* HookGetLightSourceAt(int index)`

Returns fake LightSource when `lightSourceArray[fakeLightIndex]` is indexed

## Part 4: HookActorOpaqueDraw
Hooks a call to `void ActorDrawOpaque(ModelMaterial* pModelMaterial)`

Backs up the pointer to the modelMaterial this function receives for use in Part4

## Part 5:  SetPixelShaderConstant
Here we adjust the opacity of the specular reflections depending on object/character

### Pseudo-code of the control flow with nested functions to help illustrate:
```
SilentHill2()
{
    // Hook inserts code here
    ActorDrawTop()
    {
        UnnamedFunction()
        {
            ActorOpqaueSetup()
            {
                // Hook inserts code here
                GetLightSourceCount()
                
                // Hook inserts code here
                GetLightSourceAt();
            }
            
            // Hook inserts code here
            ActorOpaqueDraw()
            {
                SetPixelShaderConstant();
            }
        }
    }
}
```


Common/Settings.h
```diff
@@ -91,7 +91,8 @@
 	visit(WhiteShaderFix, true) \
 	visit(WidescreenFix, true) \
 	visit(WndModeBorder, false) \
-	visit(WoodsideRoom205Fix, true)
+	visit(WoodsideRoom205Fix, true) \
+	visit(SpecularFix, true)
```
elishacloud, line 95: This list is mean to be alphabetized.

Common/Settings.ini
```diff
@@ -274,6 +274,9 @@ WhiteShaderFix = 1
 ; Fixes spawning placements for objects in Woodside Apartments Room 205. (0|1)
 WoodsideRoom205Fix = 1
 
+; Restores specularity to characters and objects throughout the game. (0|1)
+SpecularFix = 1
```
elishacloud, line 278: For the most part, in each section of the ini we alphabetize the items.

Patches/Specular.cpp
```diff
@@ -0,0 +1,396 @@
...
+static void Part3(ModelMaterial* pModelMaterial)
+{
+	// This function replaces a call to `void ActorDrawOpaque(ModelMaterial* pModelMaterial)`
+	// We copy off the pointer to a static variable, so we can reference it in Part 4
+
+	pCurrentMaterial = pModelMaterial;
+	ActorDrawOpaque(pModelMaterial);
+}
+
+static HRESULT __stdcall Part4(IDirect3DDevice8* /*This*/, DWORD Register, void* pConstantData, DWORD ConstantCount)
```
elishacloud, line 229: Is there a reason this function is not in the "IDirect3DDevice8.cpp" file?  It should be part of the `SetPixelShaderConstant()` function in that file.

dllmain.cpp
```diff
@@ -472,6 +472,14 @@ void DelayedStart()
 		SetSingleCoreAffinity();
 	}
 
+	FindGetModelID();
```
elishacloud, line 475: Shouldn't this be inside the `if (SpecularFix)` statement?

Patches/Patches.h
```diff
@@ -99,6 +99,8 @@ void PatchTexAddr();
 void PatchTownWestGateEvent();
 void PatchTreeLighting();
 void PatchXInputVibration();
+void FindGetModelID();
+void PatchSpecular();
```
elishacloud, line 103: This list is also alphabetized.

dllmain.cpp
elishacloud: Better yet, what if you added this line to the `PatchSpecular()` function and then you can remove the line for this out of the `patches.h` file.

Patches/Specular.cpp
```diff
@@ -0,0 +1,396 @@
...
+	// This function replaces a call to `LightSourceStruct* GetLightSourceStruct(int index)`
+	// When we hit the index 1 greater than the real light source count, return our fake
+
+	if (useFakeLight && index == fakeLightIndex)
+		return &fakeLight;
+	else
+		return GetLightSourceStruct(index);
+}
+
+static void Part3(ModelMaterial* pModelMaterial)
```
elishacloud, line 220: Can this get a better name than `Part3`?

Patches/Specular.cpp
elishacloud: Also can we get a better name for this?

elishacloud:
@FrozenFish24, Great job on this!  Nice to see this coming together.  I added a few comments for you.  Most of the updates should be pretty easy.  My biggest concern is the `Part4()` function.  I would really like to keep all the Direct3D8 functions in the wrapper section.

Patches/Specular.cpp
```diff
@@ -0,0 +1,396 @@
...
+	// This function replaces a call to `void DoActorOpaqueStuff(ModelOffsetTable* pOffsetTable, void* arg2)`
+	// Backup the materialCount and pointer to first material for use in later Parts
+
+	materialCount = pOffsetTable->materialCount;
+	pFirstMaterial = (ModelMaterial*)((char*)pOffsetTable + pOffsetTable->materialsOffset);
+
+	DoActorOpaqueStuff(pOffsetTable, arg2);
+}
+
+static int __cdecl Part1()
```
elishacloud, line 178: Is there a reason this function is not in the "IDirect3DDevice8.cpp" file? It should be part of the GetLightSourceCount() function in that file.  Can this get a better name than `Part1`?

Patches/Specular.cpp
```diff
@@ -0,0 +1,396 @@
...
+		break;
+	case ModelID::chr_item_dmr:
+		if (GetCurrentMaterialIndex() == 1) return true;
+		break;
+	}
+
+	return false;
+}
+
+static void __cdecl Part0(ModelOffsetTable* pOffsetTable, void* arg2)
```
elishacloud, line 167: Can this get a better name than `Part0`?

Patches/Specular.cpp
```diff
@@ -0,0 +1,396 @@
...
+	if (useFakeLight)
+	{
+		fakeLightIndex = lightSourceCount;
+		return lightSourceCount + 1;
+	}
+
+	return lightSourceCount;
+}
+
+static LightSource* __cdecl Part2(int index)
```
elishacloud, line 209: Can this get a better name than `Part2`?

Patches/Specular.cpp
```diff
@@ -0,0 +1,396 @@
...
+	int lightSourceCount = GetLightSourceCount();
+
+	for (int i = 0; i < lightSourceCount; i++)
+	{
+		auto pLight = GetLightSourceStruct(i);
+		if (pLight->light.Type == D3DLIGHT_DIRECTIONAL)
+			useFakeLight = false;
+
+		if (pLight->light.Type == D3DLIGHT_SPOT)
+			inSpecialLightZone = true;
```
elishacloud, line 197: We already have a variable for checking if you are in special light zone.  I wonder if that should be replaced by this?

elishacloud (changes requested):
Please make requested changes.  Feel free to push back if you don't agree with any of my requests.

Common/Settings.h
FrozenFish24: Done

Common/Settings.ini
FrozenFish24: Done

dllmain.cpp
FrozenFish24: @PolyMega [explains it here](https://github.com/elishacloud/Silent-Hill-2-Enhancements/issues/155#issuecomment-738350704), the [old method](https://github.com/elishacloud/Silent-Hill-2-Enhancements/blob/90c8cf479bdcec768126f0f50cb221e90f7c6603/Wrappers/d3d8/IDirect3DDevice8.cpp#L1054) of detecting the Woodside Room 208 TV static (based on checking texture dimensions) stops working when `IncreaseNoiseEffectRes` is active.

Using `GetModelID()` makes this really easy to solve, but it means we need to call `FindGetModelID()` even when `SpecularFix` is disabled. I could make it so it's only called by dependent fixes (Just `SpecularFix` and `EnableSoftShadows` at the moment) if you'd prefer that?

Patches/Patches.h
FrozenFish24: Done

Patches/Specular.cpp
FrozenFish24: Done

Patches/Specular.cpp
FrozenFish24: This isn't actually a call to any D3D function. A `LightSource` isn't a `D3DLIGHT8`, though it contains one. My choice of name probably doesn't make that clear, I could rename `LightSource` `SH2LightSource` or something to help clarify?

Patches/Specular.cpp
FrozenFish24: Done

Patches/Specular.cpp
FrozenFish24: Done

Patches/Specular.cpp
FrozenFish24: Good catch! I've moved this into `IDirect3DDevice8.cpp`. I now set `specularFlag` to 2 in `HookActorOpaqueDraw` so it doesn't trigger erroneously. This does add a tiny bit of complexity as there is an earlier call to `SetPixelShaderConstant()` that we don't want to touch, so we have to 'tick' `specularFlag` down and only run the code when it's equal to 1.

Patches/Specular.cpp
FrozenFish24: I'll look into this when I get a chance, I'm not entirely sure it remains valid outside `ActorDrawTop` and its sub-functions

FrozenFish24:
Thanks for the feedback @elishacloud! Will investigate inSpecialLightZone soon.

dllmain.cpp
elishacloud: > I could make it so it's only called by dependent fixes (Just `SpecularFix` and `EnableSoftShadows` at the moment) if you'd prefer that?

Yes, that sounds good.  I try to make things be able to be disabled if possible.

Patches/Specular.cpp
elishacloud: I see.  Ok then you can keep it as is.

Patches/Specular.cpp
elishacloud: Thanks.  It looks like it will always be the second call after `HookActorOpaqueDraw()`.

dllmain.cpp
FrozenFish24: How's that? I just remembered `EnableSoftShadows` has no "setup" function to put it in because it's not a patch.

elishacloud: Looks good.

FrozenFish24:
I've swapped out the use of the two `GetSpecializedLightX()` functions with `InSpecialLightZone` and it seems to work correctly. I'm not seeing any of the flickering issues we were running into either.

Sorry to bother you with more testing @Polymega, but you know this stuff better than me, would you be able to take a look at this build and make sure the special lighting zone behavior is still correct?

[d3d8.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/6026953/d3d8.zip)

You might need to add `SpecularFix = 1` to your Setting.ini or it will crash. If it works correctly I'll add code so `EnableXboxShadows` installs `HookGetLightSourceCount` and can work independently.

Wrappers/d3d8/IDirect3DDevice8.cpp
```diff
@@ -534,7 +534,7 @@ HRESULT m_IDirect3DDevice8::SetRenderState(D3DRENDERSTATETYPE State, DWORD Value
 		else // Main campaign
 		{
 			IsEnabledInRoom54 = false;
-			if (GetCutsceneID() == 0x4E || (GetSpecializedLight1() != 0x01 && GetSpecializedLight2() != 0x01))	// Exclude specialized lighting zone unless in specific cutscene
+			if (GetCutsceneID() == 0x4E || InSpecialLightZone)	// Exclude specialized lighting zone unless in specific cutscene
```
elishacloud, line 537: I think this is suppose to exclude special light zones.

Shouldn't it be:
```c++
if (GetCutsceneID() == 0x4E || !InSpecialLightZone)
```

FrozenFish24: Yep, silly typo. Now I'm not so convinced I was testing the right situations. We'll see what @Polymega says.

Polymega: Happy to help and test. I'll do this soon. With your new changes to the code, are there certain rooms/conditions in particular you can think for me to check, FrozenFish? Or just do a general sweep?

FrozenFish24: Thanks @Polymega! I think a general sweep, plus a double check of the flickery transitions we dealt with back in [this PR](https://github.com/elishacloud/Silent-Hill-2-Enhancements/pull/181) should suffice.

Polymega: I haven't done a thorough test yet, but from quick testing, [this section of code](https://github.com/elishacloud/Silent-Hill-2-Enhancements/blob/master/Wrappers/d3d8/IDirect3DDevice8.cpp#L548-L552) doesn't appear to be working now. James' self shadow should be disabled until a certain camera cut happens, then it's restored for the remainder of the cutscene.

Polymega: Also, not a big deal, but if you want to be more accurate, I'd change the variable name from `IsEnabledInRoom54` to `IsEnabledForCutscene52` or something similar. (For this code, 0x54 is for a cutscene ID and not a room ID.)

elishacloud: Ok, let's not worry about the duplicate SpecializedLight variables for now.  Just revert this back and I can reconcile these variables later, after the pull request is done.

FrozenFish24: Reverted. Thanks for the assistance @Polymega.

Polymega: No problem. If the code changes get to a good point by this weekend and you guys are ready for me, I'll sit down and do a serious playtest at that time.

elishacloud: @Polymega, below is the merged build with all of the changes from both branches.

Testing build:  [d3d8.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/6041083/d3d8.zip)

Polymega: So excited to give this a proper playthrough/testing soon! Thank you both very, very much. ❤️

Polymega: Took things nice and slow with my playthrough and everything looks great! Congratulations to you once more, FrozenFish! And thank you for the help and assistance, Elisha! 🥂

elishacloud: Nice!  I also did some testing with `InSpecialLightZone` and `GetSpecializedLight1()` and `GetSpecializedLight2()` and they are all three referring to different things.  There does not seem to be any way to rationalize these.  So it looks like we are just going to have to keep different variables for each of these.

BTW: @Polymega, does this also fix the TV shadow issue?

Polymega: >BTW: @Polymega, does this also fix the TV shadow issue?

Yes, it does!

FrozenFish24: Great to hear there's no new issues, thanks for all the help with this feature both of you!

PR #491, opened 2022-02-22
Implement FixAdvancedOptions (Fixes #370)

### LINKED ISSUE #370: Advanced Options Save Confirm Bug
@belek666 
I've found a small bug with the Advanced Options menu and wanted to ask if this is something you could look into? This issue is not critical, so if you wouldn't mind helping, please review this at your leisure:

https://youtu.be/UIOLbo_42XM

I noticed when the "Change the settings?" box appears within the Advanced Options screen, it will make the text description at the very bottom either disappear or show the wrong text string. It will also either disable or incorrectly apply the "3D effect" on certain text choices (refer to the images below).

- This appears to be a vanilla PC bug
- This happens with any language selected
- This happens with all three executables
- This does not happen on any of the other Option screens (only Advanced Options screen)

--------------------------------

Not sure if it's related to the bug, but I've noticed that if "Noise Effect" is highlighted when the pop-up box appears, it no longer has the 3D effect. If "Shadows" or "Fog" was highlighted when the pop-up box appears, then it incorrectly highlights "Advanced Filters" while the pop-up box appears (this can also be seen in the video linked above). 

Examples:

Screenshot 2021-01-27 16-39-34
Screenshot 2021-01-27 16-39-44
Screenshot 2021-01-27 16-40-00
Screenshot 2021-01-27 16-39-50

nipkownix:
Continuing from #370, almost every problematic option had its own little quirk, but I believe I got it working pretty nicely.

This PR adds the following new option:

> ; Fixes issues in the Advanced Options menu, such as incorrect descriptions appearing when the confirmation prompt shows up. (0|1)
> FixAdvancedOptions = 1

@Polymega, if you want to tweak the name/description, let me know.

Here's a test build:
[d3d8-advOptionsFix-v1.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/8117974/d3d8-advOptionsFix-v1.zip)

### DIFF
--- a/Common/Settings.h
+++ b/Common/Settings.h
@@ -30,6 +30,7 @@
 	visit(EnableTexAddrHack, true) \
 	visit(FastTransitions, true) \
 	visit(Fix2D, true) \
+	visit(FixAdvancedOptions, true) \
 	visit(FixAptClockFlashlight, true) \
 	visit(FixAudioThreadDeadlock, true) \
 	visit(FixChainsawSpawn, true) \
--- a/Common/Settings.ini
+++ b/Common/Settings.ini
@@ -108,6 +108,9 @@ MainMenuFix = 1
 ; Loads translated main menu text selections for the North American v1.0 and 1.1 EXE. 'start01*.tex' files must be present in 'sh2e\pic\etc\' for this fix to work. (0|1)
 MainMenuTitlePerLang = 1
 
+; Fixes issues in the Advanced Options menu, such as incorrect descriptions appearing when the confirmation prompt shows up. (0|1)
+FixAdvancedOptions = 1
+
 ; Disables the screen position feature in the game's options menu, which is no longer needed for modern displays. (0|1)
 LockScreenPosition = 1
 
--- a/Patches/AdvancedOptionsFix.cpp
+++ b/Patches/AdvancedOptionsFix.cpp
@@ -1,0 +1,184 @@
+/**
+* Copyright (C) 2022 Bruno Russi
+*
+* This software is  provided 'as-is', without any express  or implied  warranty. In no event will the
+* authors be held liable for any damages arising from the use of this software.
+* Permission  is granted  to anyone  to use  this software  for  any  purpose,  including  commercial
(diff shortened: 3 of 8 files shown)


Polymega:
Beautiful work here, nipkow! Improvements like this are great as it shows that extra layer of love/attention being put in to the project. Thank you very much.

I'm expecting to have something else to playtest for the project in the near future. I can try and break your fix at that time as well. I'll report back if I find anything.

Really appreciate this! It's these small pieces of polish that really add up in the long run.

nipkownix:
Nice! No need to thank me. Hopefully everything works as expected, even with you trying to break it.

Ah, something not related to this PR, but that I've been meaning to ask: why are we using 0 and 1 for booleans in the .ini file instead of ``true`` and ``false``? Not sure if there's any reason for it, but it got me curious when I was adding the new .ini option :p

elishacloud:
> Ah, something not related to this PR, but that I've been meaning to ask: why are we using 0 and 1 for booleans in the .ini file instead of `true` and `false`? Not sure if there's any reason for it, but it got me curious when I was adding the new .ini option :p

I used 0 and 1 because there is less chance of errors and it is multi-language aware.  "yes", "true", "enabled" and "on" [should all work](https://github.com/elishacloud/Silent-Hill-2-Enhancements/blob/16258da02785ae2849ccd917b302b4df8148af66/Common/Settings.cpp#L100-L108) the same as "1".  But I think that "1" will create less possible mistakes.  For example, if someone misspells "true" as "treu" it will just quietly disable the setting.

_BTW: I plan on looking at these pull requests later and responding._

nipkownix:
Ahhh, thanks for the clarification. Makes perfect sense.

No rush with the PRs, by the way.

Polymega:
Thanks, Elisha. Great to hear from you here!

Polymega:
Tested on all executables and couldn't break it. Solid work here, nipkow! 👍

elishacloud:
Nice work.  Thx!

PR #492, opened 2022-02-22
Replace "SH2PC Title" with our window title

nipkownix:
This is just a small change for something that bothers me a bit, so feel free to reject it if you don't think it is necessary.

This PR changes the game's window title form "SH2PC Title" to "Silent Hill 2: Enhanced Edition":

image

"SH2PC Title" _feels_ like a placeholder title that should have been changed before release, but... it never happened, I guess.

``ClassName`` is kept as "SH2PC Class" since something might rely on it (?) and the user never sees that anyway.

@Polymega let me know what you think. Maybe I'm just too nitpicky about minor details like this one 👀

### DIFF
--- a/Patches/Patches.h
+++ b/Patches/Patches.h
@@ -109,6 +109,7 @@ void PatchSFXAddr();
 void PatchTexAddr();
 void PatchTownWestGateEvent();
 void PatchTreeLighting();
+void PatchWindowTitle();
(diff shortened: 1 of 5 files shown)


Polymega:
Love it!! Thanks for taking the initiative!

elishacloud:
Nice work.  Thx!

PR #498, opened 2022-03-11
Setup Tool - Updater integration

nipkownix:
This PR adds a different update mode based on my Setup Tool (which we discussed via email some time ago).

Basically, if ``SH2EEsetup.dat`` exists in the game's folder, the update checking will check if there are updates available for all components, using the same .csv files that are used by the Setup Tool. If ``SH2EEsetup.dat`` doesn't exist, the already existing update method is used instead (only updating d3d8.dll).

I've done my best to keep this as streamlined as possible, but feel free to reject this and reimplement it using your own coding style if you prefer. This is at least a good way to know how the Setup Tool does things, so you know exactly how to achieve this.
If you prefer me to changed something instead, let me know.

Cheers.


elishacloud:
Thanks for putting this together!

Issue #424, opened 2021-04-17
Pausing at final boss triggers loud sound loop in credits


The-King-of-Toasters:
At the final boss, if you pause while she's in the middle of her moth attack, the moth sound will become *very loud* and loop continuously in all further cutscenes, movies, credits and the result screen. Not sure if this is a known bug (haha).

Here's me [triggering the bug](https://www.sgregoratto.me/paste/sh2/triggering-bug.mp4), and [the aftermath](https://www.sgregoratto.me/paste/sh2/feelsgoodman-clap.mp4).

Polymega:
Thanks for making us aware of this. I was able to replicate the bug. My gut says this might be a tricky/annoying one to fix. I'll add it to our to-do list.

Thanks again!

Mechahamsters:
Was a fix for this ever found? Ran into this in my very first play through of the game and it kinda ruined the ending.

Polymega:
Not yet. We're looking into properly fixing the game's infamous sound loop bug at this time. My hope is, if/when we find a fix for that, we will have a better general understanding of how the game handles audio calls and can apply that knowledge to hopefully fix this issue.

But you're right that it's important enough to fix. A Hail Mary fix would be to kill all audio threads upon exit of the final boss room ID, similar to as we've done when loading a Game Result file while in-game.

Mechahamsters:
I saw here was a 1.1 version being tested in the thread mentioned above that seemed to have good results with looping audio, although not necessarily this bug. You think it would be worth a shot on my second play through (got in water on the first)

Polymega:
I'm not sure at this time if the final boss butterfly SFX loop is the same issue as the general sound loop bug the game experiences. My gut says they're probably not related so you'd still experience it.

In the meantime the best thing to do would probably be to try and not pull up your pause menu while the butterfly attack SFX is active/playing.

Mechahamsters:
Yeah, the way that triggers, if you switch weapons it'll happen. I might still give a new playthrough a shot on that beta branch and report my findings if yall are interested.

Polymega:
I tried replicating it again and couldn't provoke it. But to be fair, the last time I tested this was serval months ago and I can't quite remember how to do it. 

But perhaps the current audio loop solution (in the current test build) fixes it? If you would want to try yourself that'd be great.

Mechahamsters:
Where can I find that test build?

Polymega:
[...]

PR #504, opened 2022-03-30
Fix for final boss's moth attack and chainsaw sound looping isuue

TheMachineAmbassador:
I've shared what i did until this time on here already: https://github.com/elishacloud/Silent-Hill-2-Enhancements/issues/424 so if i start typing maybe it can be take too long to read :)

Not so notibacle but this also fixes an issue where when you press the inventory button while in that moth room at the Westside apartment, moths chasing you would teleport to their location.

### DIFF
--- a/Common/Settings.h
+++ b/Common/Settings.h
@@ -86,6 +86,7 @@
 	visit(SaveGameSoundFix, true) \
 	visit(RowboatAnimationFix, true) \
 	visit(SetBlackPillarBoxes, true) \
+	visit(SfxSoundLoopFix, true) \
 	visit(Southpaw, false) \
 	visit(SpecularFix, true) \
 	visit(SteamCrashFix, true) \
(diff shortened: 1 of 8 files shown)


Polymega:
You're the best! Thank you so much for your terrific work! May I ask one favor? I think the average person might mistake the name of this feature for fixing the overall sound looping bug with the game (the one caused by multicore usage). I know you wrote what sounds it specifically fixes in the description, but people may just read the fix name and make an assumption.

Would you be okay with changing its name/description to something like following?

```ini
; Fixes specific sound looping issues with the chainsaw and moth attack sounds. (0|1)
SpecificSoundLoopFix = 1
```

Polymega:
If anyone can think of a different name that's short but descriptive/distinct enough for people to not associate it with the multicore sound bug, please post it!

elishacloud:
Nice work @TheMachineAmbassador!

elishacloud:
Here is the latest update with all of the fixes so far:  [d3d8.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/8399903/d3d8.zip)

PR #512, opened 2022-05-01
Gemini criware vs2022

Gemini-Loboto3:
A minor fix to ADX decoding to suppress the noise effect and changed the ini switch to be an int with 2 being no error spitting. I've also taken care of unreferenced parameters so they don't trigger VS to produce warnings.


Polymega:
Thanks a ton, Gemini!

Elisha, when this gets merged, can you please set `EnableCriWareReimplementation = 2` to be the default in the INI? Thank you

PR #514, opened 2022-05-12
Criware

Gemini-Loboto3:
Reversed 1 and 2 for debugging, set 1 to default in the ini.


elishacloud:
Nice work @Gemini-Loboto3!

Polymega:
Elisha, may I ask you to compile a build with this update? TheMachineAmbassador has mentioned that the PR may have some stability issues. I tried out a build he compiled and it consistently freezes the game when leaving the beginning bathroom. I'd like to make sure it also happens in your compile.

elishacloud:
I tested this and it seems to consistently freezes the game when leaving the beginning bathroom for me also.  I reviewed all the check-ins and all four of them have this same issue.  So the issue was caused by this check in:  fc80f1d52ada307be28cece7903d3f59364f7e63

image

Gemini-Loboto3:
Could it be a major mess up by git while merging? @elishacloud, maybe you could try and integrate the raw sources from my other dll? It's here: https://github.com/Gemini-Loboto3/sh2_tests/tree/main/criware

A direct replacement should be good, since the hooks and ini variable are in separate sources.

Unless the problem comes from Visual Studio itself when you use the xp devchain? 🤔 I used VS2019 with v142 to compile the EE test builds.

elishacloud:
> Could it be a major mess up by git while merging? @elishacloud, maybe you could try and integrate the raw sources from my other dll?

I tested this and it seems the files are the same.  Anyways, the same problem happens when I use the raw sources.

> Unless the problem comes from Visual Studio itself when you use the xp compiler?

I tried disabling XP compiler and just using v141, but it still hangs.  However, when I update to v142 it no longer hangs.   Unless there is a bug in the compiler (which I doubt), it seems like there is a bug in the code.  None of the code is compiler specific so it should work on both v142 and v141.

I don't want to move to v142 to "solve" this issue since that is just masking the bug and not a real fix.  I had a similar bug before in my code and it took a while to find it but it ultimately had to do with me not reserving a large enough buffer in one location.  But I have no idea what this bug could be.

elishacloud:
Here is the failing dll compiled with v141: [d3d8.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/8742084/d3d8.zip)

elishacloud:
Here is the working dll compiled with v142:  [d3d8.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/8742202/d3d8.zip)

_Note: both dlls are using identical code and VS settings._

elishacloud:
Here is the dll compiled with Visual Studio 2019 but using the v141 toolset.  It still hangs on this one:  [d3d8.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/8742235/d3d8.zip)

MarioTainaka:
> Here is the failing dll compiled with v141: [d3d8.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/8742084/d3d8.zip)

Just caught this in my email and decided to test it as well. and YUP this one hangs

MarioTainaka:
> Here is the working dll compiled with v142: [d3d8.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/8742202/d3d8.zip)
> 
> _Note: both dlls are using identical code and VS settings._

This one's fine.

MarioTainaka:
> Here is the working dll compiled with v142: [d3d8.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/8742202/d3d8.zip)
> 
> _Note: both dlls are using identical code and VS settings._

And this one hangs again..

MarioTainaka:
> Here is the dll compiled with Visual Studio 2019 but using the v141 toolset. It still hangs on this one: [d3d8.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/8742235/d3d8.zip)

Can confirm

Gemini-Loboto3:
It was the way I was waiting as fast as possible for threads to be in a safe state. This is caused by v141 with XP compatibility generating weird assembly. I changed the wait variable to be volatile so the compiler forces the code to re-read it, instead of just pre loading it into EAX, which causes it to keep checking the same immutable value over and over again.

That XP toolchain is the devil.

Polymega:
Neutrally asking this as someone from the outside looking in, but is XP support important to preserve?

elishacloud:
> Neutrally asking this as someone from the outside looking in, but is XP support important to preserve?

Just to be clear here.  This had nothing to do with XP support.  It would hang for _both_ `v141` and `v141_xp` toolsets.  The issue was with compiling it on Visual Studio 2017, even if XP compatibility was disabled.

Gemini-Loboto3:
I didn't use VS2017 though, just 2019 with the xp toolchain. 😛 With how often Microsoft changes compiler specs and refines optimizations to account for undefined behavior like this, that's why I suggested switching for the future.

From various documents I've read in the past, Microsoft struggled a lot to get as close as possible to the C/++ standards.

Polymega:
I see. Thanks for the info!

PR #519, opened 2022-05-20
XP compatibility

Gemini-Loboto3:
Made thread wait variable volatile because the XP compiler is dumb and generates broken assembly.


elishacloud:
Thanks @Gemini-Loboto3!

Here is the compiled dll with this update:  [d3d8.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/8744858/d3d8.zip)

PR #522, opened 2022-05-22
Inventory BGM Fix

TheMachineAmbassador:


Polymega:
Thank you, Machine Ambdassador!!

TheMachineAmbassador:
Your welcome 😊

Polymega:
Machine Ambassador,
I am both embarrassed and sorry about this... I was so concentrated on this weird BGM happening when pulling up the inventory screen it never occurred to me to check and see if this also happens when pulling up the map or pause menu.

And... yep, this issue also happens when pulling up the map or pause menu during a BGM fade. May I ask if you can look into extending this fix to also fix it for the map/pause screen?

TheMachineAmbassador:
Yep i was expecting that i was tested on catacomb and hotel map but i wasn't get any issues so i didn't add any checks for that. I will add right now :)

TheMachineAmbassador:
Here is the new release [Release.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/8758801/Release.zip) . If it's fixed i will send another PR.

TheMachineAmbassador:
No problem, i've did last changes on other PR today and i'm pretty sure that fixed our issue all the way.

Polymega:
@TheMachineAmbassador 
Can I get an emergency fix for this? I just discovered [Pianissimo Epilogue](https://www.youtube.com/watch?v=W5u1kTM3wE8) does not play on certain endings when this fix is enabled:

### `FixInventoryBGM = 0`

https://user-images.githubusercontent.com/27395260/176319291-363f3523-3cd6-4932-9786-9a6492cdb451.mp4

### `FixInventoryBGM = 1`

https://user-images.githubusercontent.com/27395260/176319377-aed07fbe-0957-4a3f-97c1-6ac5d011880c.mp4

It does not play for UFO, DOG, Maria, and In Water endings. Here is a save folder of the main campaign endings: [Folder 01.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/9006017/Folder.01.zip)

TheMachineAmbassador:
Oh boy! I'm looking at right now.

Polymega:
One other oddity I found with `FixInventoryBGM`: There are a few points in the game where, if you pull up the Load Game screen, the music will either stop while this screen is opened or will incorrectly play a track that shouldn't be heard yet:

### Music stops playing here when it should keep playing
https://user-images.githubusercontent.com/27395260/182244949-327691f2-1d9a-47f4-ba08-3fb5d9635882.mp4

### Music starts playing here when it shouldn't
To make this happen, create a save right before meeting Maria in the park. Load this save, then pull up the Load Game screen.

https://user-images.githubusercontent.com/27395260/182245605-3f6f0c46-0087-471c-8c5a-9fb90973e54e.mp4

There may be other spots, but these are two that I found which behave incorrectly while `FixInventoryBGM = 1`.

Please let me know if you'd prefer me to open a new ticket for this.

Thank you!

Polymega:
I should also note that, as far as I can tell, this only happens with the Load Game menu and _not_ the Save Game menu.

TheMachineAmbassador:
> Please let me know if you'd prefer me to open a new ticket for this.

Hi @, for now we don't need to open a new ticket. I will check what can i do about the issue when i get back to home. I have some ideas to fix it.

TheMachineAmbassador:
Well i checked the issue right now and as you know fixing the issue for the load screen was a bit awkward because same screen used on the main menu. When we force this 01d7ead8 memory address to 15 at that some points on game it fixes our problem but as you said it brokes other sound stuff. 

sound

That memory address that i sended you gets own value by this 00515a40 function so i will do some reverse engineering in here and see what we can do about it. When we touch something on this game it brokes other thing, it's a miracle this game can be playable on latest windows :).

Polymega:
> When we touch something on this game it brokes other thing, it's a miracle this game can be playable on latest windows :).

Without exaggeration, I'd say nearly everything we have done for this project directly or indirectly relates to the Save/Load anywhere feature in some way. Please don't be discouraged... we have worked many miracles in the past and, with your help, we  will continue to make more miracles!

Polymega:
> fixing the issue for the load screen was a bit awkward because same screen used on the main menu

Can you refresh my memory: Why was this an issue again? I'm trying to remember why this was a concern in the hopes of coming up with a solution.

Polymega:
So, I've only found two instances of this issue so far and these could be the only two issues for this throughout the entire game. I'd need to do an extensive test for this, but I'm going to guess this is a rather isolated issue. One thing we could do is add Room ID exclusions?

So something like:

```c++
if (GetRoomID() != 0x01) {  /* Is not beginning bathroom */ 
	if (EventIndex > 3 && EventIndex < 10 || EventIndex == 0x10 || MenuEventIndex == 0x11)
		{
			__asm
			{
				jmp jmp_return
			}
	}
}
```

I'd need to come up with a solution for the "meeting Maria" one later on.

What do you think?

TheMachineAmbassador:
Well we can do that but we have to find stage by stage whole bug points :). Doing that might solve the issue quickly but we can't be sure the issue is fixed or not on whole stages.

Polymega:
True, but I also don't want you pulling out your hair out over this. If you'd like to look into this deeper to make a global, generic fix then I certainly won't stop you! _But..._ I don't mind sitting down and testing all areas of the game for this bug. If my gut feeling is correct, this bug may only affect a small handful of areas and, if so, we could just do a simple Room ID exclusion check.

TheMachineAmbassador:
Nah not a problem, today i don't have nothing to do so i can look at the code's much deeply if i can't fix today we can try that method you said to me :).

TheMachineAmbassador:
Hi @Polymega, when you have time can you please test it, i fixed the issue i guess.

[Release.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/9275018/Release.zip)

Polymega:
Unfortunately, if you quickly go to the Load Game screen while a previous audio track is fading away, [it resurrects the original issue that we've fixed](https://user-images.githubusercontent.com/27395260/168485868-1ab8bd21-24b5-4ef0-860b-6ccab96db66b.mp4).

(^ Pretend I went to Load Game instead of Inventory in that video.) 

Let me do a long test and see how many rooms/areas this affects.

Polymega:
Can you remind me what the underlying conflict with the Load Game screen is again? Was it something to do with the Load Game screen and the main menu sharing the same Event Index value?

TheMachineAmbassador:
> Can you remind me what the underlying conflict with the Load Game screen is again? Was it something to do with the Load Game screen and the main menu sharing the same Event Index value?

Yep main menu and in game load save menu shares same event index and main menu events index value.

We can find buggy point's with this 01d7ead8 memory address actually, i don't know the reason but i think when that addresses value is zero that issue occurs.

Polymega:
From doing a deep dive into this, pulling up the Load Game menu is changing the music in more areas than I was hoping. I tested up to the apartments before deciding there are too many spots to try and fix this on a per-area basis. 

If I had to take a guess, it seems like, when this does happen, the game calls and will play _the next track_ in an audio list.

For background music, SH2 plays ADX and AIX tracks. ADX are single-file audio tracks. AIX is a container file that holds multiple ADX tracks, so that the game can play/overlap multiple tracks at the same time. [See here for a list/example.](https://github.com/elishacloud/Silent-Hill-2-Enhancements/blob/master/AudioScripts/BGM/BGM-Map.csv)

Let's use bgm_115.aix as an example. This is audio that plays while exploring the town. This AIX contains the following ADX files within it:

- bgm_115_0000.adx
- bgm_115_0001.adx
- bgm_115_0002.adx
- bgm_115_0003.adx
- bgm_115_0004.adx

Let's say your in a certain area of the town, so the game has loaded and is currently playing the following tracks:

- bgm_115_0000 
- bgm_115_0001. 

When you pull up the Load Game screen, I want to say it is adding +1 to the audio tracks playing, so now it might be playing:

- bgm_115_0000
- bgm_115_0001
- bgm_115_0002

When you exit the Load Game menu, it will stop playing the "+1" track. This is just an uneducated theory on what's happening. What's a bit aggravating about this is that it doesn't happen all the time; it only happens in certain areas throughout the game. If it happened everywhere it would probably be more feasible to fix...

--------------------------

So.... with all that said... this is getting a lot more tricky. If you'd prefer to leave the fix as-is I fully support you on that decision.

--------------------------

> Also if you have save file of that spot, can you please share me :).

Here is a general save folder:
http://www.igotaletter.com/temp/sh2pc/General_Saves.zip

If you want to test this, I have a better spot for you: Load the "Forest" save near the top. Meet Angela in the cemetery then test _anywhere_ in the forest _after_ meeting Angela. Test in the cemetery after meeting Angela all the way to the gate under the bridge (that brings you into town).

In the cemetery, if you pull up the Load Game menu the music here will fade away. Leave the cemetery area and, anywhere else in Forest area, if you pull up the Load Game menu a new track will start to play that shouldn't be playing. Listen carefully and you'll know when the new track is playing.

Test this using the public release build (with the public `FixInventoryBGM` fix).

Here are the Forest/Beginning Town files for you to reference:
http://www.igotaletter.com/temp/sh2pc/bgm_115_tracks.zip

Polymega:
> Yep main menu and in game load save menu shares same event index and main menu events index value.

Wait... if this is the only thing that is causing this issue then surely we could find another address to help us with this, correct? I would be shocked if there aren't other addresses in the game to signify when the Load Game menu is pulled up (that are different from a value used for the main menu).

So could something like this fix it?

```c++
if (EventIndex > 3 && EventIndex < 10 || EventIndex == 0x10 || MenuEventIndex == 0x11 && ??????? == 1)
```

???????? would be the new address we'd need to find.

Could it be as simple as this?

nipkownix:
> I would be shocked if there aren't other addresses in the game to signify when the Load Game menu is pulled up

Even if there aren't, we could still create our own ``bool`` that is set to true once the load game menu is pulled up, I assume.
(Haven't been following this discussion, so this is just based on the last couple of messages I've read :p)

TheMachineAmbassador:
Hi @Polymega, i've checked every possibilies i can do but i guess we have to find all buggy point areas in the game😞. Actually there is another function that i can reverse engineer but it's like reverse engineering save encryption algorithm, it's just confusing a bit.

> Even if there aren't, we could still create our own bool that is set to true once the load game menu is pulled up, I assume.

Unfortunately 😢 we're already checking is load game menu pulled up or not with the "MenuEventIndex == 0x11".

Polymega:
Let's not worry about this any longer then! This issue isn't game-breaking or anything, so we shouldn't stress too much about it. Thank you _very_ much for looking into this as much as you did. Sorry it was a pain in the butt!

nipkownix:
So, these issues you were showing in the videos are caused by ``FixInventoryBGM``? What was that option fixing, btw? This PR doesn't have a description, so I'm a bit lost :p

Polymega:
[Check out this portion of the video.](https://youtu.be/LylafxQta9w?t=683)

This is what `FixInventoryBGM` fixes. If you pull up just about _any_ menu (inventory, map, save screen, load screen, options) while a BGM track is fading away, it will incorrectly play another track while the menu is pulled up.

Machine Ambassador found a fix for this. It works great overall except, the other day I found that, with this fix, the BGM track might incorrectly stop playing or start layering another BGM track on top while in the Load Game menu. [See this post here.](https://github.com/elishacloud/Silent-Hill-2-Enhancements/pull/522#issuecomment-1201715340)

This is only happening with the Load Game menu. All other menus (inventory, map, save screen, options, etc) are fine; just the Load Game menu exhibits this issue. 

If the issue is solely because we currently can't differentiate between when the Load Game menu is pulled up vs. the main menu then we should be able to find a new address to use. However, going off what Machine Ambassador has been saying, it sounds like it is not that simple and this bug is a result of a deeper, underlying issue.

This issue/bug with the Load Game menu isn't game-breaking or anything, so it's not the end of the world if we can't address it. And, as it currently stands, I think it's better to keep the Load Game menu as part of `FixInventoryBGM` for what it _does_ fix for the Load Game menu.

Polymega:
PS - A useful page for bookmark: https://github.com/JokieW/SilentHillDatabase/blob/master/SH2/SH2%20PC%20memory%20addresses

Using that page, we can look at `FixInventoryBGM` in the module:
https://github.com/elishacloud/Silent-Hill-2-Enhancements/blob/master/Patches/PatchInventoryBGM.cpp#L22

And find the following:
- MenuEventIndex: https://github.com/JokieW/SilentHillDatabase/blob/master/SH2/SH2%20PC%20memory%20addresses#L36
- EventIndex: https://github.com/JokieW/SilentHillDatabase/blob/master/SH2/SH2%20PC%20memory%20addresses#L37-L51

```c++
	if (MenuEventIndex == 13 /*[normal gameplay]*/ || MenuEventIndex == 17 /*[load game menu]*/  && EventIndex == 0 /*[load game menu]*/)
	{
		if (EventIndex == 11 /*[game result screen]*/)
		{
			*muteSound = 0xF;
			__asm
			{
				jmp jmp_return
			}
		}
		
		if (EventIndex > 3 && EventIndex < 10 /*[just about every type of menu]*/ || EventIndex == 16 /*[pause menu]*/ || MenuEventIndex == 17 /*[normal gameplay]*/)
		{
			__asm
			{
				jmp jmp_return
			}
		}
	}
	__asm
	{
		jmp jmp_to_loop
	}
}
```

Issue #209, opened 2020-02-04
Always play game load/save sound in respective menus.


Polymega:
Something maybe for a future update, which mimics SH3 PC's approach to it:

The "game load" sound effect only plays to full completion when starting a new game/loading a game from the main menu. The game load sound does play, but will get cut off short if you load a save from the PC pause menu.

The "save sound" will only play when interacting with a red save square, and will not play at all if you choose to save your game from the PC pause menu. 

It would be a nice touch to restore these sounds, and have them play fully, even when saving/loading from the PC pause menu, as what SH3 PC does.

Polymega:
@elishacloud @bigmanjapan

Setting 0249C8B8 byte to 1 while in-game will provoke the "game load" SFX, but it only loops the first second infinitely while the value is 1. 0249C8BC byte also seems to be linked/associated with 0249C8B8.

If I find other addresses which I think might be relevant I'll edit this post or make a new one.

Polymega:
@TheMachineAmbassador
Thanks for taking an interest here. :)

I hope it'd be a relatively straightforward solution of finding the instructions that call the save sound when interacting with a red square and using these instructions when triggering the save menu through the PC pause menu.

You'll just want to ensure that what you're changing is for when you pull up the save screen via the PC pause menu vs. interacting with a red square.

If it helps, we already have a variable made for [00932178](https://github.com/JokieW/SilentHillDatabase/blob/master/SH2/SH2%20PC%20memory%20addresses#L36) called `EventIndexAddr` that you can see here: 
- https://github.com/JokieW/SilentHillDatabase/blob/master/SH2/SH2%20PC%20memory%20addresses#L36-L50

TheMachineAmbassador:
I guess i found interesting thing in this memory 0x48c1d0. When you trigger red square save thing you trigger this function and this trigger function calls sound manager function 0x48c301.
![sound_id](https://user-images.githubusercontent.com/77938614/155433748-cc18dc90-3a27-4d03-8094-8343954373d0.png)
If you change this push 2743 (2743 is sound id i guess. e.g. 2744 is door open id) it changes sound. Now i'll search for the pause menu sound manager and if i compare menu buton index with save game buttons id i can change sound id with the red square save sound's id or maybe i can find more interesting things in pause menu save function that doesn't need to change any id :)

TheMachineAmbassador:
[...]

PR #526, opened 2022-05-24
Some little changes and extended fix for other Pause menu's

TheMachineAmbassador:


TheMachineAmbassador:
I've updated source with much more clear one. Here is the build [Release.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/8764330/Release.zip).

Polymega:
I didn't do another complete playthrough but I tested all of the trouble spots. I tested pulling up the inventory screen, map screen, and pause menu. Everything worked great under these circumstances. 

However, if you pull up the pause menu _and very quickly_ go to the Options or Save/Load screen, the BGM bug will still happen on these screens. It will also happen if you go to the Memo screen very quickly from the Inventory screen. I think to fix this you would just need to add 

```c++
EventIndex == 0x0 || EventIndex == 0x7 || EventIndex == 0x8 || EventIndex == 0x9
```
[to this line](https://github.com/elishacloud/Silent-Hill-2-Enhancements/pull/526/commits/95ae921636a6d651f9a714919eff0c1222e06fc4#diff-8819b32f1a13dbb2c4a1f2485f180bd1da9a0b19ab7ca7725f89bb49c69cee70R18), correct?

TheMachineAmbassador:
Yep that is correct also According to this [link](https://github.com/JokieW/SilentHillDatabase/blob/5ea589bffb5b33cee634a711f0ddc1e8683d16b8/SH2/SH2%20PC%20memory%20addresses#L37) we can tell that function i patched is if game is in load screen then send to loop sound, if not then change the sound or something like that. If it's so i don't need to add all game states :) i can simply write if game is in load screen then go to loop but i have to think something different for the "load save" because it seems there is no id of that. I can simply add save state's id to that.

Polymega:
So I might be oversimplifying this or maybe missing something, but couldn't you apply this fix globally without any conditional checks? Or does that break something?

If we need the conditional checks you _might_ want to keep using `EventIndex` only because we now have all events properly covered now. Using the load screen trigger might not be as fool-proof?

TheMachineAmbassador:
> So I might be oversimplifying this or maybe missing something, but couldn't you apply this fix globally without any conditional checks? Or does that break something?

This condition's are used for to keep bgm playing while you are in loading screen, the reason of that bug is game waits to check if player succesfully passed through any door and then slowly fade-out's music but when you press inventory button game look at's the code and see there is only 0x4 check in there so it continues the play music :) because game doesen't exactly stop music when you pass through door it still loops sound and waits until music completely stop. you guy's already did fadeout's much more quick already so we can remove condition it just makes much more quickly fade out music.
 
> If we need the conditional checks you _might_ want to keep using `EventIndex` only because we now have all events properly covered using it. Using the load screen trigger might not be as fool-proof?

I tried to save state address but you're right it didn't work 😀 so if we want to our condition met for "Load Game" we must find something different or we can remove conditions. If you want to i can send you without an condition build.

Polymega:
> This condition's are used for to keep bgm playing while you are in loading screen [...] you guy's already did fadeout's much more quick already so we can remove condition it just makes much more quickly fade out music. [...] so if we want to our condition met for "Load Game" we must find something different or we can remove conditions

If you are okay with doing this, let's keep using the `EventIndex` check and add `0x0, 0x7, 0x8, 0x9` to the list. I suggest this as we already know this approach works and all we'd need to do is add those events in the list to fix all remaining problems.

Polymega:
Damn. I'm not at home to check, but I'm going to assume that `EventIndex 0x0` is used for both the Load Screen and probably the Main Menu?

Polymega:
Whatever you think works best and covers all bases! (Inventory, map, pause, options, memo, save/load)

All I'd ask is that the fade out time of the BGM tracks remains roughly the same as it currently is. People will easily notice if the fade time of the audio is noticeably shorter or longer.

Believe it or not, we haven't touched the fade time of the BGM tracks. It probably feels faster since we've shortened the _visual_ fade time of the game (aka [FastTransitions](http://enhanced.townofsilenthill.com/SH2/config.htm#LightingTransitionFix)).

TheMachineAmbassador:
Well i tested with 2 different code and yep it's noticeably faster. Thanfully we have ps2 code in our hand if pc and ps2 versions handles this states same we can use that.

TheMachineAmbassador:
> Where all does `EventIndex 0x0` trigger? I know it triggers while on the Load Screen, but does it also trigger on the Main Menu? Does it trigger elsewhere such as during a new room load?
> 
> Sorry to ask but I won't be home for a while to check myself.

Well according to my reverse reverse engineering data, it just resets stats on game, i guess there might be a bug  on "Load Game" function that does EventIndex to 0
event

TheMachineAmbassador:
Yep i found it 0x04028c2 this instructor does event index to 0, when i nopped this it makes EventIndex to 0x10 correctly.

elishacloud:
Currently we use DirectSound to stop the audio when in game results.  See [here](https://github.com/elishacloud/Silent-Hill-2-Enhancements/blob/7f7d411b68a20d34c20438e6548929fec105d2cf/Wrappers/dsound/IDirectSoundBuffer8.cpp#L363).  I wonder if we could use this code to stop the audio in game results?  This is a better way to stop the sound than I am using.

Polymega:
> Yep i found it 0x04028c2 this instructor does event index to 0, when i nopped this it makes EventIndex to 0x10 correctly. But i don't know is there a reason of game's setting that Addr to 0

Does this _only_ set the **Load Screen** to 0x10? Or does it set **all** instances where it would normally be 0x0 to 0x10?

If it only sets the Load Screen to 0x10, would the instructions change the value back to 0x0 upon selecting a save file to load?

TheMachineAmbassador:
> Currently we use DirectSound to stop the audio when in game results. See [here](https://github.com/elishacloud/Silent-Hill-2-Enhancements/blob/7f7d411b68a20d34c20438e6548929fec105d2cf/Wrappers/dsound/IDirectSoundBuffer8.cpp#L363). I wonder if we could use this code to stop the audio in game results? This is a better way to stop the sound than I am using.

I've tested it without using that and our new function, there was no issue. Maybe new criware patch might solve this? Even if there is issue, yeah this new patch should solve this because i'm patcing this global Bgm point.
function

> Does this _only_ set the **Load Screen** to 0x10? Or does it set **all** instances where it would normally be 0x0 to 0x10?

And i tested it too with NOP'ing that instructor and there is no issue. This function 004484e0 clears save data before the loading/saving game already so we can NOP that. Also if we want to be sure yeah we can place EventIndex thing to before loading save.

> If it only sets the Load Screen to 0x10, would the instructions change the value back to 0x0 upon selecting a save file to load?

Pause menu array does EventIndex value to 0. It would be 0x10 normally but even i don't know why game sets EventIndex to 0. First i thought game needs to clear value but then i set breakpoint to here 00401fe3 to see is game really clears data but breakpoint never triggered, i guess nothing needs that. Also yeah it turning back 0x0 to after we load game.

TheMachineAmbassador:
Also it's not related with this issue but i found the solution of this while researching: https://github.com/elishacloud/Silent-Hill-2-Enhancements/issues/209

00515562 - this was the that stops sound after loaded game. If you nop this and load a save it won't stop until ends :)

Polymega:
> 00515562 - this was the that stops sound after loaded game. If you nop this and load a save it won't stop until ends :)

Oh snap! From the _very little_ testing I've done this seems to work just fine? And the extra wait time between loading save files is very minimal, but it makes the whole presentation a lot nice since the sound now plays to completion.

We should test this further soon!

Polymega:
I spoke too soon, it stops BGM on a new load but does not stop SFX that play for a long time or dialogue files. These are BGM sounds that are technically dialogue:

- The TV static in the apartments
- The beginning lake trail (after leaving the Observation Deck)
- Any place with water dropping (such as the Labyrinth or Alternate Hotel)
- Any place where it rains (after the transition to the Alternate Hospital or the final boss fight)

(I'm probably forgetting some.) Since these are considered dialogue files, they do not stop playing on a new file load.

------------------

> And i tested it too with NOP'ing that instructor and there is no issue. This function 004484e0 clears save data before the loading/saving game already so we can NOP that.

Nice! So it should be safe to make this change to prevent the wrong BGM track issue for all scenarios then?

TheMachineAmbassador:
> I spoke too soon, it stops BGM on a new load but does not stop SFX that play for a long time or dialogue files. These are BGM sounds that are technically dialogue:
> * The TV static in the apartments
> * The beginning lake trail (after leaving the Observation Deck)
> * Any place with water dropping (such as the Labyrinth or Alternate Hotel)
> * Any place where it rains (after the transition to the Alternate Hospital or the final boss fight)
>  (I'm probably forgetting some.) Since these are considered dialogue files, they do not stop playing on a new file load.

Not a problem i will work on that, important thing is we found where the issue causes :)


> Nice! So it should be safe to make this change to prevent the wrong BGM track issue for all scenarios then?

Well ye.... no  😥That is true it's safe to NOP that value but unfortanetly but unfortunately issue is still there because i've set a breakpoint on to my code and noticed, "load game" function works different then "save game", when you press load game function it completly freezes everything in game but save game function doesen't. Also i noticed this right now [link](https://github.com/JokieW/SilentHillDatabase/blob/5ea589bffb5b33cee634a711f0ddc1e8683d16b8/SH2/SH2%20PC%20memory%20addresses#L36) become 0x11 when we are into loading screen but on other things the value is 0xD. But anyway i just wanted to scare you with technical things :)) if you noticed that picture i sent elisha above there is a main_menu_things value there and after pressing load game button main_menu_thing is become 0x11 so our code never reaches to that  so i've changed that :) and added main_menu_things to our comparison. Now it's working.

TheMachineAmbassador:
Hello @elishacloud, i've added this line to your utils code (same code with SearchAndGetAddress)
util but before i wanted to ask you is there any code smilar you added smilar to this. I was gonna use pattern search but somehow it was crashing game.

TheMachineAmbassador:
Right now it seems i fixed our problem and also we can assume "Main menu events index" value as "load save" id. Because in game load game button using same function as main menu.
[Release.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/8780293/Release.zip)

Polymega:
Thank you! I did some quick tests and the wrong BGM track will still play if you quickly pull up the Memo screen (from the Inventory) while the previous track is still fading.

TheMachineAmbassador:
Because i didn't add Memo screen's index value 😂 Now it's added [Release.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/8782951/Release.zip)

elishacloud:
> Hello @elishacloud, i've added this line to your utils code (same code with SearchAndGetAddress)

It looks like all you are doing here is subtracting the ByteData?  If so, you don't need a whole new function for that.  No need to modify the utils code.  You can just send it a negative number.  See [here](https://github.com/elishacloud/Silent-Hill-2-Enhancements/blob/7f7d411b68a20d34c20438e6548929fec105d2cf/Patches/DelayedStart.cpp#L44) as an example.

> I was gonna use pattern search but somehow it was crashing game.

Pattern search is MUCH slower than SearchAndGetAddress.  If I used pattern search for every code search in the module it would likely take at least 20 additional seconds to start the game.  I got the additional loading time less than 1 second with SearchAndGetAddress.  I only included pattern search because it was needed for the widescreen fix code that I import.

Polymega:
Yep! Works great now! And could this line be further simplified?

```c++
if (EventIndex >= 4 && EventIndex <= 10 || MenuEventIndex == 0x11)
```

Polymega:
> if (EventIndex >= 4 && EventIndex <= 10 || MenuEventIndex == 0x11)

I am terrible at programming btw, but I think I can see an issue with my suggestion of using `=` in this line. Since you are using `&&` and not `||` logically it wouldn't make sense to have it equal more than one value? Or is that allowed?

TheMachineAmbassador:
> It looks like all you are doing here is subtracting the ByteData? If so, you don't need a whole new function for that. No need to modify the utils code. You can just send it a negative number. See [here](https://github.com/elishacloud/Silent-Hill-2-Enhancements/blob/7f7d411b68a20d34c20438e6548929fec105d2cf/Patches/DelayedStart.cpp#L44) as an example.

Thank you, that is much more better i'll update code again :)
 
> Pattern search is MUCH slower than SearchAndGetAddress. If I used pattern search for every code search in the module it would likely take at least 20 additional seconds to start the game. I got the additional loading time less than 1 second with SearchAndGetAddress. I only included pattern search because it was needed for the widescreen fix code that I import.

Ah I see, we can tell you're method much more modern way to find addresses :). Usually people uses patterns for doing mods but this EE project is much more complicated than the others. When i'm looking at other project i never saw their mods continued, they usually stops crash issues and then stops progeressing project. 

> I am terrible at programming btw, but I think I can see an issue with my suggestion of using `=` in this line. Since you are using `&&` and not `||` logically it wouldn't make sense to have it equal more than one value? Or is that allowed?

Nah what you say makes sense actually but In real-life conversation, we could say like this "If EventIndex is greater than or equalt to 4 and If EventIndex is smaller than or equal than 10" you said, we can tell it's true in way but as you see other EventIndex value is 0x10 not 10 :) 0x10 in decimal means 16 so if we set EventIndex to 0x10 that covers the other indexes like FMV,Game Over Screen etc.

elishacloud:
Looks good!  Let me know if it's ready to be merged.

Polymega:
> 0x10 in decimal means 16 so if we set EventIndex to 0x10 that covers the other indexes like FMV,Game Over Screen etc.

Mark my words: Things like 0x10 being 16 will be something I will always, always forget. 😅

TheMachineAmbassador:
> Mark my words: Things like 0x10 being 16 will be something I will always, always forget. 😅

Next time i'll write in decimal, i promise 😂

> Looks good! Let me know if it's ready to be merged.

I can tell it's ready, what do you think @Polymega ?

Polymega:
> I can tell it's ready, what do you think @Polymega ?

I didn't do another full playthrough but I tested all the spots I know where this issue happens and it works great! I feel confident this solution is solid. 😊

elishacloud:
Nice work @TheMachineAmbassador!

elishacloud:
Here is the new build with all updates so far:  [d3d8.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/8792305/d3d8.zip)

PR #538, opened 2022-06-28
Fix: Sound doesen't play at specific endings

TheMachineAmbassador:
Kids don't forget to add return after work is done.

### DIFF
--- a/Patches/PatchInventoryBGM.cpp
+++ b/Patches/PatchInventoryBGM.cpp
@@ -24,6 +24,10 @@ __declspec(naked) void __stdcall FixInventoryBGMBugASM()
 		if (EventIndex == 0xb)
 		{
 			*muteSound = 0xF;
+			__asm
+			{
+				jmp jmp_return
+			}
 		}
 		
 		if (EventIndex > 3 && EventIndex < 10 || EventIndex == 0x10 || MenuEventIndex == 0x11)


TheMachineAmbassador:
Release added also it includes beta quick save text thing so don't forget to replace build back after test is done.

[Release.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/9006176/Release.zip)

elishacloud:
Thanks @TheMachineAmbassador.

@Polymega, here is the fix added to the release candidate: [d3d8.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/9006198/d3d8.zip)
 
Once tested I will merge the PR.

Polymega:
This looks to fix things up! Thank you Machine Ambassador for the quick turnaround on this!

@elishacloud 
I finished my full playthrough and everything else was good. Thanks!

@TheMachineAmbassador
> also it includes beta quick save text thing so don't forget to replace build back after test is done.

Sounds good! I will play around with this soon and email you back. 😄

Polymega:
Oh, Elisha, unrelated: But could you share with Machine Ambassador what code in the module you use to check the current resolution the game is using?

elishacloud:
> But could you share with Machine Ambassador what code in the module you use to check the current resolution the game is using?

Sure.

@TheMachineAmbassador, the current resolution of the game is stored in the variables [here](https://github.com/elishacloud/Silent-Hill-2-Enhancements/blob/8632c72103f28d319a152525f6bdb0bd554a0d37/Wrappers/d3d8/d3d8wrapper.h#L47).

They are set by this line of code [here](https://github.com/elishacloud/Silent-Hill-2-Enhancements/blob/8632c72103f28d319a152525f6bdb0bd554a0d37/Wrappers/d3d8/IDirect3D8.cpp#L337-L338).

Also, you can checkout the code [here](https://github.com/ThirteenAG/WidescreenFixesPack/blob/f3ef32b1856f27bd8d28cb01a1e4aaebd64ec115/source/SilentHill2.WidescreenFix/dllmain.cpp#L92-L98) in the Widescreenfix, as this is internally where the game stores the screen size.

TheMachineAmbassador:
Thank you @elishacloud, i guess my problem is related to hud stuff,

hud

I'm currently rewriting the game's render text function but unfortunately the widescreen fix doesn't affect this function because HUD variables are at the inner scope of the function. I'm trying to put the text in the left corner but after changing resolution it's sliding right because of aspect ratio. I could've simply compared the aspect ratio and changed the text position by that but I don't know it is a good idea.

elishacloud:
> I'm trying to put the text in the left corner but after changing resolution it's sliding right because of aspect ratio.

[Here](https://github.com/elishacloud/Silent-Hill-2-Enhancements/blob/c36369101b76b6438bcfca075f7be776098936fd/Patches/Resolution.cpp#L270-L289) is the code that gets run every time the resolution changes.  You could change this code to update the HUD ratio also.

TheMachineAmbassador:
Well yeah but can't we do that general, i mean whole rendering stuffs for the fonts in game is rendering in 4:3. For the better explanation.


In this photo i've set x position of both text's to 0 and as you see text is still at 4:3.
16 9redbox
While i was writing you i've already noticed what i have to do already and what you trying to tell me, i guess all i have to do is just search the 4:3 red box (layer) that i draw :) and find the data of that and then change that values to with your code. Well right now i have to look for some start point i guess 😄

Polymega:
Might be a silly question but, does positioning accept negative values? If so, could you use a negative value to offset its horizontal position when using ratios wider than 4:3?

TheMachineAmbassador:
Yep it does and i can use but i have to add comparisons for every aspect ratios and it wont gonna be used by other font render functions. So we have two ways to go.  (Like i did for all fixes  😂)

1-) I can add comparisons for every aspect ratio (We can say it's esaiest way )

2-) It might took a bit but we can try find layer that for the fonts and it will affect all texts in the game.(Like now loading text,subtitles etc)

Polymega:
> i can use but i have to add comparisons for every aspect ratios

You may not need to. Could you use just the resolution variables Elisha provided and use the current resolution height to determine this?

Let's say the person is playing in 1280x720. 
Calculate height of 720 * 1.3333333333333 = 960 (4:3 aspect ratio at this height)
960 - 1280 (current width used by game) = -320
-320 + 50 (padding example so text isn't touching edge) = Offset horizontal text by -270

Polymega:
> and it wont gonna be used by other font render functions.

This is probably be a good thing. The "regular" text is, for all intents and purposes, displaying fine as it currently is. Repositioning the GAME SAVED/CAN'T SAVE text would be all that's needed.

TheMachineAmbassador:
Well game calc doesen't work like that unfortunately i'm trying to find different formula but for now but i already found different way. I simply copy pasted this code [link](https://github.com/elishacloud/Silent-Hill-2-Enhancements/blob/c36369101b76b6438bcfca075f7be776098936fd/Patches/Resolution.cpp#L167 ) and editted a bit. The result is like this, 
[Release.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/9028935/Release.zip) I still don't know is it support 21:9 but better that nothing i guess. All i did was in this build subtracting fixed 20 value with 30.

Polymega:
> I still don't know is it support 21:9 but better that nothing i guess

Unfortunately, it looks to have a limit of ~16:9.

Screenshot 2022-07-01 09-11-47

> Well game calc doesen't work like that unfortunately

I'm not meaning to come across as stubborn or anything when I ask this, but I figured the formula/calculation would be done on our end through the module? 

- [BufferHeight](https://github.com/elishacloud/Silent-Hill-2-Enhancements/blob/8632c72103f28d319a152525f6bdb0bd554a0d37/Wrappers/d3d8/d3d8wrapper.h#L47) * 1.333333333 = A
- A - [BufferWidth ](https://github.com/elishacloud/Silent-Hill-2-Enhancements/blob/8632c72103f28d319a152525f6bdb0bd554a0d37/Wrappers/d3d8/d3d8wrapper.h#L47)= B
- B + (BufferWidth * 0.05) = C

`C` is final horizontal offset of text
`(BufferWidth * 0.05)` would be 5% padding from left so text isn't touching edge. (% value could be changed to whatever.)

Now, if the positioning values used for this text are not 1:1 to the resolution values, then that would throw a wrench into this. ...unless we can determine a conversion for it to do at the very end? 🤔  In any case, the idea with the above formula is that you would not need to hard code any positions depending on aspect ratio used.

TheMachineAmbassador:
The calc formula is correct but thing is we don't know which coordinate system text at. We can substract text with any value to slide it left but it's not goes left like that, the build that build i sended you now is also includes text positions already :). When i substracting with value that we found i'm not actually subtracting with 0 because the text doesn't slide left with that, I have to substract with 4095 in the build you will find the reason. But with the changes a bit i did with you're formula i guess we're a bit close to fix pos of text.

[print_pos.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/9029540/print_pos.zip)

Polymega:
>  If you ask me why i substracted with 4095, it's just 0 point for the pos.

Ah, now the math is getting really confusing for me. So at 4:3 it uses a value of 30, and subtracting with 4095 is used as it's the origin point (0) for its position?  😵

... but still, there ought to be a way to understand how the game is using these values for positioning and convert it in a way we better use? May I ask if you could create a Cheat Table where I can alter this position live?

Oh, I used an absolutely ridiculous aspect ratio to test and the current build looks to still have limits:

Screenshot 2022-07-01 10-34-27

TheMachineAmbassador:
You can search value of position in this build [Release.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/9030590/Release.zip) . You can change value by simply addding new memory address on cheat engine and put the hello value to Address. Data type is short.

PR #594, opened 2022-09-03
Debug and Info Overlay in DirectX 8

mercury501:
Debug and Info Overlay, on by default, enable with respectively CTRL + D and CTRL + I.


elishacloud:
Nice work @mercury501!

PR #602, opened 2022-09-14
Overlay Refactor

mercury501:
Now draws each overlay with one DrawText call.


Wrappers/d3d8/Overlay.cpp
```diff
@@ -155,118 +177,129 @@ void Overlay::DrawDebugOverlay(LPDIRECT3DDEVICE8 ProxyInterface)
...
+	{
+		DebugFont->DrawTextA(FontStruct.String, -1, &DropShadowRect, FontStruct.Format, TextColors.Black);
+		DebugFont->DrawTextA(FontStruct.String, -1, &FontStruct.Rect, FontStruct.Format, FontStruct.Color);
+	}
+}
 
-	TextStruct.String = OvlString.c_str();
-	TextStruct.Rect.top += TextSpacing;
-	TextStruct.Rect.bottom += TextSpacing;
+void Overlay::DrawMenuTestText(LPDIRECT3DDEVICE8 ProxyInterface, Overlay::D3D8TEXT FontStruct)
```
elishacloud, line 242: Can you rename this function `DrawMenuTestText()` to something more descriptive?  Maybe just remove the word "Test" from it?

elishacloud:
Other than the two comments I added to the code, I found an issue with this code.  If I start the game at 1360x768 and then turn on both overlays and then switch to 800x600 then the info overlay disappears.

Wrappers/d3d8/Overlay.cpp
mercury501: Of course!

mercury501:
I renamed the function and added a check for changed resolution.
Thanks for spotting the bug!

elishacloud:
I just realized you called this "MenuTest" all over the code. I wonder if it makes sense to call this "InfoMenu" or something like that instead?

mercury501:
Reverted last name change!

mercury501:
Should be good now!

elishacloud:
Thanks @mercury501!

PR #606, opened 2022-09-23
Changes to the Spanish translation for the Launcher

IlDucci:
- Summarized the two descriptions that were overflowing. The Fullscreen Image Scaling description is currently the longest one used in the Launcher, and while I have summarized it, I would still need to expand the description box one extra line (the largest section is the Misc one, and has space for that extra line). We could also think about this as localization-proofing that particular description, as I'm sure Spanish will not be the only language that needs that much space (not without an important loss of information).
 - Adding translations for the two new strings that have been added. What do you people think about making the language change question a double-language one?

### DIFF
--- a/Launcher/config_es.xml
+++ b/Launcher/config_es.xml
@@ -42,7 +42,7 @@
 
       <Feature name="FullscreenImages">
         <Title>Escalado de imágenes a pantalla completa</Title>
-        <Description>Incorpora la compatibilidad para el Image Enhancement Pack (paquete de mejora de imágenes) del proyecto para que se muestre correctamente. Es necesario activar esta opción si se utiliza el Image Enhancement Pack. El escalado funciona tanto para las imágenes originales a pantalla completa como para las mejoradas.&#x0a;&#x0a;Automático: selecciona automáticamente entre las opciones de ajustar a imagen o llenarla según la relación de aspecto, dando prioridad a la mejor presentación.&#x0a;Ajustar a área de visualización: muestra las imágenes en pantalla completa sin recortarlas. Podrían aparecer bandas o columnas negras según la relación de aspecto.&#x0a;Llenar área de visualización: escala las imágenes a pantalla completa para que llenen el área de visualización, eliminando cualquier banda o columna negra, cubriendo una relación de aspecto máxima de 16:9.</Description>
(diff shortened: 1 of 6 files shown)


Polymega:
> We could also think about this as localization-proofing that particular description, as I'm sure Spanish will not be the only language that needs that much space (not without an important loss of information).

Sure. How can we go about doing that?

Also, you could change [the double line break](https://github.com/elishacloud/Silent-Hill-2-Enhancements/pull/606/commits/12b0432007e55790ab9caa93c52c17b8879f350a#diff-6fee42b58b1c93527d9d534b9c1bb83f24becc07dd0a024959d6f0cbc4dcfc8cR45) (`&#x0a;&#x0a;`) to a single line break (`&#x0a;`) to give yourself an extra line.

> Adding translations for the two new strings that have been added. What do you people think about making the language change question a double-language one?

Elisha will correct me if I'm wrong, but I believe this string's text is taken from the _currently loaded_ XML file. So, if in the future, five languages are offered for the Config Tool, you'd need to include all five translated lines [here](https://github.com/elishacloud/Silent-Hill-2-Enhancements/blob/master/Launcher/config.xml#L1096) within every translated XML file. I think the three options here would be:

- Poll the other language's XML file just for this string. (Ideal solution)
- Include multiple translated sentences in this pop-up box. (Messy)
- Have this string display in the _currently_ selected language. (As it currently does)

IlDucci:
> Also, you could change [the double line break](https://github.com/elishacloud/Silent-Hill-2-Enhancements/pull/606/commits/12b0432007e55790ab9caa93c52c17b8879f350a#diff-6fee42b58b1c93527d9d534b9c1bb83f24becc07dd0a024959d6f0cbc4dcfc8cR45) (`&#x0a;&#x0a;`) to a single line break (`&#x0a;`) to give yourself an extra line.

I can, for sure, but it looks a bit wall-of-texty.
imagen

> Sure. How can we go about doing that?

I was referring to expanding the description box the size of one line upwards:

imagen

That is also why I was referring to the Misc tab, as it has the highest height in terms of options used.

elishacloud:
> That is also why I was referring to the Misc tab, as it has the highest height in terms of options used.

Good idea.  I'll make this a bit larger so it could fit one more line of text.

> Maybe you can put most of the misc tab options under game fixes?

I don't think we need to move these around unless there is a good reason.  I would rather organize the options on the tab than move them around just to make space...

Launcher/config_es.xml
```diff
@@ -1093,6 +1093,8 @@
     <S id="PRG_Launch_exe">sh2pc.exe</S>
     <S id="PRG_Ini_name">d3d8.ini</S>
     <S id="PRG_Ini_error">No se pudo guardar el archivo d3d8.ini.</S>
+    <S id="PRG_Lang_confirm">¿Deseas cambiar de idioma al %s?&#x0a;&#x0a;Do you want to switch to language %s?</S>
```
elishacloud, line 1096: This line won't work.  The string can only have a single `%s` in it otherwise it will crash.  Maybe we should remove the English text from here?

IlDucci: Whoopsie-daisy. Fixed now.

elishacloud:
Thanks @IlDucci!

Issue #647, opened 2022-10-28
Post server migration: Report issues of Setup Tool here


Polymega:
Server migration is done and distribution for the project's file packages has been changed to a more stable solution.

If you run into any troubles still with the Setup Tool, please report them here.

Balthazor44:
Something I just noticed, I ran the game and got a message that an update was available and I chose the option to close the game and run the setup tool. I let it install the update, but now when I run the game, it still tells me that an update is available, even though the setup tool says there isn't.

Polymega:
The next time you launch the Setup Tool you should receive a message that the tool itself needs to be updated. Let it update and see if that fixes the update loop issue.

Balthazor44:
I ran the setup tool and it did update, but unfortunately that just made things stranger. Now when I run the game, it still tells me an update is available, but now when I choose the option to run the updater I get a screen telling me it's downloading SH2EEsetup_New.exe, then it launches the configuration tool.

nipkownix:
Seems to be some logic problem we missed during testing. Sorry about that. I'll take a look at the issue tomorrow, but you can safely ignore the update dialogue for now.

Balthazor44:
> Seems to be some logic problem we missed during testing. Sorry about that. I'll take a look at the issue tomorrow, but you can safely ignore the update dialogue for now.

Okie dokie, let me know if there's anything else you need from me.

Moonwalwer:
Im having a problem here, im dowloading this for the first time, when i run the setup tool and  start installing it says "runtime error at 308:1926 out of range, what do i do?

nipkownix:
Thanks! Could you please upload here your "SH2EEsetup.dat" file? Should be in the game's folder.

Polymega:
@nipkownix 
For this particular "out of range" error I suggested they restart their computer and it fixed the problem. [See ticket here.](https://github.com/elishacloud/Silent-Hill-2-Enhancements/issues/651) I Google searched this error and there wasn't many results, but one of the very few pages had someone suggest rebooting their PC. 😅

Polymega:
@Balthazor44 
At your earliest convenience, can you please upload the `SH2EEsetup.dat` file found in the game's main directory? Just click + drag the file onto into the text field here when you respond. Thanks!

Balthazor44:
[...]

PR #654, opened 2022-10-29
AutoUpdate: Additional logging

nipkownix:
Log which packages the module thinks need updates, and their versions.

Could help when diagnosing things like this: https://github.com/elishacloud/Silent-Hill-2-Enhancements/issues/647#issuecomment-1295988838


elishacloud:
Thanks @nipkownix

PR #685, opened 2022-12-17
Add short delay before fade-in transitions

Murugo:
This PR introduces a hidden bool option `DelayedFadeIn` which adds roughly 2/15 seconds of delay before fading in the screen. This delay helps to hide some distracting animation bugs that occur for a few frames after room transitions. These include [Maria interpolating from her last animation](https://github.com/elishacloud/Silent-Hill-2-Enhancements/issues/28#issuecomment-1324452978) in the previous room into an idle pose, and pools of blood forming quickly on the ground.

This is intended to be a visual change only, i.e. James is still able to move around during this delay. It's possible that there is other game logic that ties player control to whether the screen is fully faded in. Hopefully, the difference is unnoticeable since the delay is only a fraction of a second.

See attached video for a comparison.

https://user-images.githubusercontent.com/49109252/208234775-c99632c7-ab20-4127-86fe-c03425867b4e.mp4

### DIFF
--- a/Common/Settings.h
+++ b/Common/Settings.h
@@ -13,6 +13,7 @@
 	visit(ClosetCutsceneFix, true) \
 	visit(CreateLocalFix, true) \
 	visit(d3d8to9, true) \
+    visit(DelayedFadeIn, true) \
 	visit(DisableCutsceneBorders, true) \
 	visit(DisableEnlargedText, true) \
 	visit(DisableGameUX, true) \
@@ -176,6 +177,7 @@
 	visit(CustomFontCol) \
 	visit(CustomFontRow) \
 	visit(CustomModFolder) \
+    visit(DelayedFadeIn) \
 	visit(DisableLogging) \
 	visit(EnableDebugOverlay) \
 	visit(EnableInfoOverlay) \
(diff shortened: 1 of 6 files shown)


Polymega:
Beautiful work here, Murugo. Your work and attention to detail is great. Thank you so much for creating this feature and it was a pleasure to test it out for you.

(Sorry, I fat-fingered the review request on my phone. You can only add emojis to posts when viewing GitHub through desktop mode on your phone!)

elishacloud:
Great job!  Everything looks good.  My only comment is that this is a marked as a hidden feature.  Is it meant to be hidden or do we want to add it to the SH2EEconfig tool?

Polymega:
> My only comment is that this is a marked as a hidden feature. Is it meant to be hidden or do we want to add it to the SH2EEconfig tool?

Murugo also asked me about this, wondering if it's worth making toggleable. I didn't think it was but I told him that decision is ultimately up to him, and he agreed it's not vital enough to make toggleable. To give you a visual example, it adds _the slightest_ of delays to fade ins but, in doing so, fixes (by hiding) different visual quirks during the fade in:

https://user-images.githubusercontent.com/27395260/210899608-6783a9dc-58ff-4787-8320-6a47fcc4e9e3.mp4

^ His fix is so solid that framerate doesn't affect the time, either!

elishacloud:
Ok, great.  This is enabled by default and there's probably not much reason to disable it.

PR #693, opened 2023-01-12
Fix subtitles during FMVs to draw on top of noise grain

Murugo:
This change introduces a boolean option `FmvSubtitlesNoiseFix` which draws subtitles on top of noise grain during FMVs.

Typically, subtitles and other text during normal gameplay are drawn on top of the noise filter. The only exception is during FMVs in which the noise filter appears on top of subtitles. This causes the font to appear slightly darker than usual (see below). This patch ensures that subtitles are always drawn on top to be consistent with gameplay text.

Thanks @Polymega for pointing this out!

https://user-images.githubusercontent.com/49109252/212016130-b386369f-f3c0-4069-94b8-e2471c571125.mp4


Polymega:
Always knocking it out of the park!

elishacloud:
Thanks @Murugo!

Here is the full merged build with all changes in it:  [d3d8.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/10453510/d3d8.zip)

PR #694, opened 2023-01-14
Tweaks and fixes for the latest Mouse input and 60 fps implementations

mercury501:
Changed places of a setting, and a fix for the toggle sprint feature

PR #700, opened 2023-01-30
Launcher - Spanish translation update

IlDucci:
Adding the latest updates as requested here: https://github.com/elishacloud/Silent-Hill-2-Enhancements/pull/694#issuecomment-1407795029 .


elishacloud:
Thanks @IlDucci!

Here is the latest build with all the changes in it:  [d3d8.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/10541375/d3d8.zip)

PR #705, opened 2023-02-04
[Launcher] Updating Spanish translation

IlDucci:
Abbreviating the LoadD3d8FromScriptsFolder string and also mentioning the folders in plural, rather than using a singular form.


elishacloud:
Thanks @IlDucci!

Here is the updated merged build:  [d3d8.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/10609321/d3d8.zip)

PR #708, opened 2023-02-18
Updated italian translation

mercury501:


elishacloud:
Thanks @mercury501!

Here is the merged build:  [d3d8.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/10774809/d3d8.zip)

PR #714, opened 2023-02-20
Config xml changes

Polymega:
Added warning lines to some feature descriptions for the Config Tool. For example: If `UseCustomExeStr` is disabled it'll also disable `DynamicResolution`. Disabling `UseCustomModFolder` will also disable `DynamicResolution` as this feature now pulls `r_menu_*.res` files from the `sh2e` path.

### DIFF
--- a/Common/Settings.ini
+++ b/Common/Settings.ini
@@ -164,7 +164,7 @@ DisableEnlargedText = 1
 ; Loads translated main menu text selections for the North American version 1.0 and 1.1 executables. Note: start01*.tex files must be present in "sh2e\pic\etc\" for this feature to work.
 MainMenuTitlePerLang = 1
 
-; Restores all language options (English, French, German, Italian, and Spanish) as selectable choices in the game's Options menu and creates language-specific pause menu text for the North American executables of the game. Note: .\sh2e\resources\r_menu_*.res must be present in the game's directory for this feature to work.
+; Restores all language options (English, French, German, Italian, and Spanish) as selectable choices in the game's Options menu and creates language-specific pause menu text for the North American executables of the game. Note: .\sh2e\resources\r_menu_*.res must be present in the game's directory for this feature to work. Note: If disabled, various other features will also be disabled. This feature should never be disabled.
 UseCustomExeStr = 1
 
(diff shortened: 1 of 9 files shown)


Polymega:
I changed the language names to be written in their native tongue, but these special characters might not be allowed for this part? https://github.com/elishacloud/Silent-Hill-2-Enhancements/pull/714/commits/97ecaf61a4146c28b502bea5ed2c2c166574d870

elishacloud:
Yeah, I just checked and it does not handle those chars correctly.  Let me see if I can figure out why.

image

elishacloud:
Here is the merged build:  [d3d8.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/10796337/d3d8.zip)

MarioTainaka:
@Polymega Dunno if it's to late to offer any suggestions but it might be wise to include an option to enable or disable redirection from the language folder maybe  leave it enabled by default. removing or renaming the folder isn't exactly elegant

elishacloud:
Yes, good idea.  It is too late for this release.  But I made a note of it for the next release.

Polymega:
I am now home to test this. When switching to the Brazilian Portuguese language I receive the following error: 

image

... and then am no longer able to launch the tool (since it's trying to launch in that language).

elishacloud:
There is something wrong with the XML in that config file.  I'll have to figure out what it wrong with it later.  For now you can go to the "HKEY_CURRENT_USER\SOFTWARE\Konami\Silent Hill 2\sh2e" registry key and delete the "LauncherLanguage" value/data.

Polymega:
Interestingly, if I copy that XML as-is to `SH2EEconfig.xml` the tool will run. I might have missed something in one of the VS files that builds the launcher...

elishacloud:
Ok, I put a fix in for this.  Here is the updated, final build:  [d3d8.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/10798848/d3d8.zip)

Polymega:
Sorry I missed that. Everything is running well now. Thank you, Elisha.

PR #752, opened 2023-03-10
Fix for quit menu horizontal navigation on DC

mercury501:
Just a small offset change.

Cheers!


elishacloud:
Thanks @mercury501!

Issue #733, opened 2023-02-25
60 FPS: Sound Glitch for footsteps when running


wesker265:
In 60 FPS mode, there is a brief audio glitch when James begins running. It does not happen when walking in 30 or 60 FPS mode.

The footstep sound should go tap ..... tap .... tap .... tap ....tap .... tap (Like in 30 FPS)

But it goes tap....tap-tap-tap....tap .... tap .... tap .... tap (in 60 FPS)

This is most apparent in open areas like on the streets and not so much in closer quarters like the first floor apartment

Polymega:
Hmm, I am not hearing this. 

https://user-images.githubusercontent.com/27395260/221385826-2cdcae86-2f9d-425e-8450-22bfdee25f1e.mp4

Are you able to record video/audio of it happening?

MarioTainaka:
> In 60 FPS mode, there is a brief audio glitch when James begins running. It does not happen when walking in 30 or 60 FPS mode.
> 
> The footstep sound should go tap ..... tap .... tap .... tap ....tap .... tap (Like in 30 FPS)
> 
> But it goes tap....tap-tap-tap....tap .... tap .... tap .... tap (in 60 FPS)
> 
> This is most apparent in open areas like on the streets and not so much in closer quarters like the first floor apartment

Does this happen on dirt surfaces as well? for example during the first section when you encounter the line figure? Does it only happen when walking or when you turn?

Polymega:
I know that, if you time it _just right,_ you can make him "stutter step" whenever James goes from walking to running. I also think, on rare occasion, this can also happen when he transitions from the beginning of his run into full sprint (which he can do while in outdoor areas). The timing of the animation change can cause that to happen, which I'm pretty sure can happen on all versions of the game.

wesker265:
For me, it's either from walking to running, or full stop to running (with
running by default enabled instead of walking). It happens on the street
outside the trailer on Saul street, and pretty much any street.

It did not happen on the first floor of the apartment.

Polymega:
Is it what you hear near the end of the video below? If so, that's normal behavior exhibited in all versions of the game. As previously mentioned, it's the transition of his movement from one animation to another. 

https://user-images.githubusercontent.com/27395260/221388090-8a07b0ff-992d-42e8-bac4-b986a4ab57ba.mp4

wesker265:
No, I'll try to take a video of the issue later tonight.

[...]

PR #761, opened 2023-03-17
Fix "double" footstep sounds for player characters

Murugo:
This PR adds a patch which removes "double" footstep sounds that can sometimes occur when playing at 60 FPS. This patch is grouped with other high framerate fixes enabled with `SetSixtyFPS`. See https://github.com/elishacloud/Silent-Hill-2-Enhancements/issues/733 for more context.

If timed correctly, it is possible to trigger an additional footstep sound in between left and right footsteps. This happens when the player changes to a different walk/run animation at the point when the previous walk/run animation loops. This is easiest to trigger outdoors -- by pressing D-pad up from an idle position, James/Maria will trigger the sound twice: once when entering a full sprint, and again when slowing down after running out of stamina. Demo:

https://user-images.githubusercontent.com/49109252/225818962-a86846e3-f09d-489d-bbce-3f1fd6f100f6.mp4

Build for testing: [d3d8.dll.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/10949882/d3d8.dll.zip)

### DIFF
--- a/Patches/DoubleFootstepFix.cpp
+++ b/Patches/DoubleFootstepFix.cpp
@@ -1,0 +1,119 @@
+/**
+* Copyright (C) 2023 Murugo
+*
+* This software is  provided 'as-is', without any express  or implied  warranty. In no event will the
+* authors be held liable for any damages arising from the use of this software.
+* Permission  is granted  to anyone  to use  this software  for  any  purpose,  including  commercial
+* applications, and to alter it and redistribute it freely, subject to the following restrictions:
+*
+*   1. The origin of this software must not be misrepresented; you must not claim that you  wrote the
+*      original  software. If you use this  software  in a product, an  acknowledgment in the product
+*      documentation would be appreciated but is not required.
+*   2. Altered source versions must  be plainly  marked as such, and  must not be  misrepresented  as
+*      being the original software.
+*   3. This notice may not be removed or altered from any source distribution.
+*/
+
+#define WIN32_LEAN_AND_MEAN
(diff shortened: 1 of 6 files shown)


Polymega:
Knocking it out of the park as always, Murugo! Thank you dearly. In the near future we will be pushing out a hotfix for the project to fix some pointer issues with the Director's Cut binary that led to crashes and removing external runtime dependencies for Xidi. This hotfix will also include your 60 FPS fix here. I remember you also expressed interest in wanting to compare and possibly sync the movement speed between keyboard and joystick. Is this something you'd still like to research? If so, we're happy to wait a bit longer before pushing out this update--not saying that to rush you, though! Just let us know!

Murugo:
Thanks @Polymega!

> I remember you also expressed interest in wanting to compare and possibly sync the movement speed between keyboard and joystick. Is this something you'd still like to research? If so, we're happy to wait a bit longer before pushing out this update--not saying that to rush you, though! Just let us know!

I started looking into this and it turns out there may be more nuances with player movement than I originally thought (e.g. comparing joystick vs. D-pad vs. WASD and PC vs. PS2). I am thinking of opening a separate ticket for discussion, but in the meantime I don't think this needs to block the hotfix.

Polymega:
> I am thinking of opening a separate ticket for discussion, but in the meantime I don't think this needs to block the hotfix.

Sounds good. Thank you very much again!

elishacloud:
Thanks @Murugo!  Nice work.

PR #828, opened 2023-08-12
Move quick save canceling fixes to a separate feature

Murugo:
This PR moves the following fixes from `GameLoadFix` (#778) to a new feature called `QuickSaveCancelFix`:

- Cancel an in-progress quick save before entering another room.
- Cancel an in-progress quick save when interacting with a save point.
- During a save, write sh2pc.sys and sh2pc??.dat on the same frame.

This change is primarily for speedrunners that prefer to disable `GameLoadFix`. With the new feature enabled, they can take advantage of quick save exploits while avoiding situations that could softlock the game or corrupt in-game text if they were to quick save at the wrong time.

`QuickSaveCancelFix` is currently a hidden feature, but PLMK if you prefer that I add this to "Misc > Game Saves" in the config tool.


elishacloud:
Thanks @Murugo.

Issue #786, opened 2023-05-14
Mary's moth attack: moths are rendered behind transparency effect of broken glass


matheusvb3:
Around the stage where you fight Mary there are many windows with broken/damaged panes which use transparent textures. When the moths are in front of these panes, the transparency effect gets rendered on top of them.

<details>
  <summary>Images</summary>
  
![moth1](https://github.com/elishacloud/Silent-Hill-2-Enhancements/assets/98937378/34f4aaba-b8cb-4286-be74-c972a1a08f6f)
![moth2](https://github.com/elishacloud/Silent-Hill-2-Enhancements/assets/98937378/133b35f7-731f-4637-9fa1-04b1f09dd595)
![moth3](https://github.com/elishacloud/Silent-Hill-2-Enhancements/assets/98937378/d6f01098-0226-4ac0-b221-dfbd5b94f8e0)
![moth4](https://github.com/elishacloud/Silent-Hill-2-Enhancements/assets/98937378/22164fb0-79c9-41e6-a368-042c6e9115bb)
![moth5](https://github.com/elishacloud/Silent-Hill-2-Enhancements/assets/98937378/5dd4fae5-d82f-41aa-ab8c-5bf9f2b7b497)
![moth6](https://github.com/elishacloud/Silent-Hill-2-Enhancements/assets/98937378/dbcf7c6c-d0b4-409d-81e5-500529e07fbb)
![moth7](https://github.com/elishacloud/Silent-Hill-2-Enhancements/assets/98937378/e5e74f1d-9ee4-4831-8148-08b7d1729897)
![moth8](https://github.com/elishacloud/Silent-Hill-2-Enhancements/assets/98937378/6de74481-011d-47e9-a381-f7fc08b0a8d3)
  
</details>

<details>
  <summary>Random rambling</summary>
  
[...]

PR #865, opened 2023-11-08
Fix moths rendering under translucent objects such as glass

Murugo:
Adding a new game fix which resolves a visual bug with moths. These objects are exclusive to Wood Side Apartments Room 202 and the final boss room. Moths now correctly draw on top of translucent meshes such as glass and certain parts of the final boss room floor. This issue was reported in #786.

MothDrawOrder1
MothDrawOrder2


elishacloud:
Nice work @Murugo!  Thanks!

PR #867, opened 2023-11-12
Sync subtitles to FMV framerate

Murugo:
Adding a new game fix `FmvSubtitlesSyncFix` which adjusts the playback speed of subtitles to match the Bink video framerate. 

Subtitles use frame numbers to determine display start/end times. These always assume 29.97 FPS. This means that an FMV with a framerate of 59.94 FPS having twice the number of frames will cause subtitles to play back twice as fast. The fix recalculates the frame number used to advance subtitles to sync correctly with the video (in this case, `adjustedFrame = frame * (29.97 / 59.94) = frame * 0.5`).


elishacloud:
Nice work.  Thanks @Murugo!

Issue #841, opened 2023-09-14
`GameLoadFix` causes CTD when quick loading within certain area


Polymega:
Don't ask how I stumbled across this; pure happenstance.

With `GameLoadFix` active, if you Quick Save in this general area of Town East **at night** and then Quick Load back in, the game will crash to desktop:

![image](https://github.com/elishacloud/Silent-Hill-2-Enhancements/assets/27395260/e2ed7c5f-0b91-400e-af01-a4f1c5d289e1)

https://github.com/elishacloud/Silent-Hill-2-Enhancements/assets/27395260/cab73c03-faa4-4966-97c7-0dab9d906615

Through tests, this bug seems to have introduced itself in [v2.1.2070.0](https://github.com/elishacloud/Silent-Hill-2-Enhancements/releases/tag/v2.1.2070.0). [v2.1.2050.0](https://github.com/elishacloud/Silent-Hill-2-Enhancements/releases/tag/v2.1.2050.0) and earlier this does not happen.

Looking at the release notes for [v2.1.2070.0](https://github.com/elishacloud/Silent-Hill-2-Enhancements/releases/tag/v2.1.2070.0), the things relating to quick saving are these:

> - Fix "Game Save" location for ultrawide screens
> - Update the fix for Quick Save text fading too quickly
> - Fix an issue where when you delete the quick save then try to load it you can no longer exit the game

I'll try and test a little bit more on my own to see if I can determine its cause, but I might not be able to, so I want to make others aware.

Also, [if you refer to this PR](https://github.com/elishacloud/Silent-Hill-2-Enhancements/pull/828), Murugo has split some of the features that were in `GameLoadFix` into their own, new feature. He sent me a test build of this change and I can confirm those features (now as `QuickSaveCancelFix`) are _**not**_ the cause of this bug. So it's likely one of those things in the three bullet points listed above.

Polymega:
> Fix an issue where when you delete the quick save then try to load it you can no longer exit the game

I can confirm this is the cause. [Here's the portion of code for that.](https://github.com/elishacloud/Silent-Hill-2-Enhancements/commit/10a78a8df771830349dd6eb7d2abaf4b08281ec8) Below is our previous discussion on why we did this:

----------------------------------

- Load a save file so you're in-game
- Press Esc > Save Game
- Find the Quick Save save file, press Backspace on it, and delete it
- Return to in-game
[...]

PR #879, opened 2023-12-05
Adjustments to GameLoadFix to resolve softlocks/crashing

Murugo:
This PR contains two bug fixes for `GameLoadFix`:

1. Fix an issue where quick loading into certain areas causes the game to crash (#841). As mentioned in the issue, the culprit is [this commit](https://github.com/elishacloud/Silent-Hill-2-Enhancements/commit/10a78a8df771830349dd6eb7d2abaf4b08281ec8).

    Originally, after deleting the quick save and attempting to quick load, the player would no longer be able to quick to the main menu from the pause screen. This happens because a flag that indicates that a load operation is in progress (`00932034`) stays enabled even after the quick load fails. The current fix NOPs the instruction that enables this flag during the load operation. Unfortunately, this change has other side effects that prevent quick loading from functioning properly. The updated fix waits for "NO QUICK SAVE" to appear on screen before resetting the flag.

4. Fix the address used to detect whether Maria (NPC) is dying to prevent the player from pausing or quick saving (#778). The correct address to use is `01FB0F4C`. I had mistakenly used `01FB0F4B`, which is set to 1 whenever Maria takes any damage or is inside an enemy hitbox.

    @Polymega pointed out that letting the hospital chase cutscene play out prevents the player from pausing or quick saving indefinitely until the player closes the game. This is because Maria is in close proximity with PH during the cutscene, and `01FB0F4B` stays at a value of 1 even after the cutscene ends.

Test build: [d3d8.dll.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/13510237/d3d8.dll.zip)


Polymega:
Thank you, Murugo!!

elishacloud:
Thanks @Murugo!

PR #881, opened 2023-12-09
Updating Spanish translation for the Launcher

IlDucci:
- Removing a dupe of the EnableMasterVolume section.
- Adding translations for all the missing strings.
- Being a bit cheeky with EnableLangPath's description to indicate users that there are two versions of the Spanish translation around: the Neutral (which is included in EE) and Castilian (for Spanish users, outside of EE's base package).

### DIFF
--- a/Launcher/config_es.xml
+++ b/Launcher/config_es.xml
@@ -362,15 +362,6 @@
         </Choices>
       </Feature>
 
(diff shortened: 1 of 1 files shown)


elishacloud:
Thanks @IlDucci!

Issue #418, opened 2021-04-09
Teddybear in hospital locker room bug


Arthandas:
James keeps looking at it as if it was a pickup item, even after taking out the needle.

Polymega:
I checked the other versions and he also does that there. I'm not saying it's "right," but rather something that's inherent to the game. I'm going to consider this one a non-issue. I don't feel it's important enough to worry about.

PR #893, opened 2024-01-16
Fix to stop James from looking at the teddy bear after taking the bent needle

Murugo:
Adding a game fix (`TeddyBearLookFix`) which prevents James from looking at the teddy bear in the hospital 2F women's locker room after picking up the bent needle.

James will look at certain objects in the room that are in his line of sight. Usually, James will stop looking at an object after he picks it up. The teddy bear is a special case since the bear remains on the table after James takes the bent needle. This can confuse players into thinking there is an additional action to take with the bear after watching the cutscene (e.g. #418).

<details><summary>Before -> After</summary>

i_bear_before

i_bear_after

</details>

### DIFF
--- a/Common/Settings.h
+++ b/Common/Settings.h
@@ -123,6 +123,7 @@
 	visit(SpecularFix, true) \
 	visit(SteamCrashFix, true) \
 	visit(SwapLightHeavyAttack, true) \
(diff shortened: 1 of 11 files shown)


elishacloud:
Thanks @Murugo!

Issue #735, opened 2023-02-26
Old Man coin is visible before it should be


gobbins:
The Old Man coin outside the apartments is visible, even before using the canned juice in the garbage chute. There are clearly two 'states' for this, as picking up the coin removes it from view.

d3d8.log and screenshots are attached.

Thank you!

![Desktop Screenshot 2023 02 26 - 03 41 39 06](https://user-images.githubusercontent.com/126386088/221403773-c2682d6f-834d-4dd5-a189-141177877b53.png)
![Desktop Screenshot 2023 02 26 - 03 42 35 06](https://user-images.githubusercontent.com/126386088/221403775-9f86f6af-fccd-4285-9b37-87806fe0475c.png)
![Desktop Screenshot 2023 02 26 - 03 42 47 35](https://user-images.githubusercontent.com/126386088/221403777-ee9dce33-cf10-45af-9048-424467c4e42f.png)
[d3d8.log](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/10833403/d3d8.log)

TheMachineAmbassador:
Hello @Polymega 😊, im started working on this bug. Will share you the details as soon. For now i know this bug is connected with this memory address sh2pc.exe+1B7A9EC. Well i was already checked what this memory is on ps2 dump and it says this is named as "game_flag". But i know this bug is not complex at all. I bet it's a logic issue in some code. Something like if not equals to some stuff show coin or something like that 😂.

Polymega:
@TheMachineAmbassador 
It's really good to hear from you again! When you have a moment, I sent you an email.

TheMachineAmbassador:
No problem, you can send me any time 😌

Polymega:
Great! I sent you the email. Let me know if it came through or not.

Polymega:
Our wonderful team member @Murugo found an easy fix for this, so it will be addressed for the next project update.

Also, since I was in here, I made the garbage look nicer. (If that isn't an oxymoron, I don't know what is.)

![Screenshot-2024-02-03-11-10-25](https://github.com/elishacloud/Silent-Hill-2-Enhancements/assets/27395260/09de65cf-3946-421c-b139-39cf33248ce4)

PR #912, opened 2024-02-07
Add fix for the Old Man Coin appearing too early

Murugo:
Adding a patch (`OldManCoinFix`) that prevents the Old Man Coin from appearing in the garbage chute until after James uses the canned juice in the Wood Side Apartments second floor laundry room (#735).

PLMK if you think this should be added as its own game fix in the config tool.


elishacloud:
Nice.  Thanks @Murugo!

PR #930, opened 2024-03-05
Extend the fog area behind the observation deck

Murugo:
Adding `ObservationDeckFogFix` under Graphics > Fog. This patch widens the fog area behind the observation deck at the start of the game to fill a 16:9 screen ratio.

This patch requires a modified **bg\ob\obGB.fcl** that increases the fog particle count from 450 to 1,575 in proportion to the new width of the fog area. To support this change, the patch also increases the maximum number of allowed fog particles from 500 to 2,000. If `ObservationDeckFogFix` is disabled while the modified .fcl file is in use, the fog particle count will be clamped to 500. Compared to 450, this should not make a significant difference.

[sh2-ob-fog.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/14491091/sh2-ob-fog.zip) contains a test build, as well as the modified obGB.fcl.

<details><summary>Before + After</summary>

opening-fog-before

opening-fog-after

</details>


elishacloud:
Nice work @Murugo!  Thanks!

PR #947, opened 2024-04-13
Fix alignment issues for the boards that are used in the BFaW tombstone riddle

FrozenFish24:
Thanks to @Polymega for providing the color-coded test textures.

Before:
image

After:
image


elishacloud:
Thanks @FrozenFish24!

PR #949, opened 2024-04-14
Fix broken UI in DC and NA 1.1 due to gravestone boards fix

FrozenFish24:
The address in the fild instruction was only valid for NA 1.0, I have also switched to patching the function call which removes the need to suppress warnings.


elishacloud:
Thanks @FrozenFish24!

PR #951, opened 2024-04-15
Fix flashlight position during the shower room drain cutscene

Murugo:
Adding a patch that prevents the spotlight from James's flashlight from suddenly shifting on the last frame of the shower room drain cutscene in Brookhaven Hospital (`ShowerRoomFlashlightFix`). Added as a hidden feature since this is a minor fix, but PLMK if this should be its own game fix in the config tool.

https://github.com/elishacloud/Silent-Hill-2-Enhancements/assets/49109252/d634fc2d-7d76-47c3-8c6b-b35fc7b48503

<details><summary>Explanation</summary>

The issue happens on frame 574 of the cutscene (the freeze frame when the "I got an elevator key" text appears). `FUN_0047AC50` retrieves the position of the active spotlight in the scene. At instruction `0047AC59`, there is a check for whether a cutscene is currently playing (`00932194 & 0x40`). If this flag is true, the function copies James's flashlight position from `01FB7D38` ("JamesLightPosOriginal"). Otherwise, it copies from `01FB7D18` ("JamesLightPos").

sh-store-spotlight

Since the cutscene effectively ends on frame 574, the cutscene flag resets on this frame, causing the flashlight position to jump ("JamesLightPosOriginal" -> "JamesLightPos").

The patch checks if the current room is the hospital shower room, and if so it checks a different flag for whether a cutscene is active, namely the status of James's object entry (`01FB1005 & 0x20`). This flag remains active on frame 574, since control hasn't yet returned to James until the player dismisses the textbox.

</details>

Test build: [d3d8.dll.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/14976388/d3d8.dll.zip)

Thanks to Bigmanjapan for researching this bug and providing addresses.

### DIFF
--- a/Common/Settings.h
+++ b/Common/Settings.h
@@ -118,6 +118,7 @@
 	visit(SetBlackPillarBoxes, true) \
 	visit(SetSixtyFPS, true) \
 	visit(SetSwapEffectUpgradeShim, false) \
+	visit(ShowerRoomFlashlightFix, true) \
 	visit(Southpaw, false) \
 	visit(SpecificSoundLoopFix, true) \
 	visit(SpecularFix, true) \
@@ -241,6 +242,7 @@
 	visit(ResX) \
 	visit(ResY) \
 	visit(SetSwapEffectUpgradeShim) \
+	visit(ShowerRoomFlashlightFix) \
 	visit(SmallFontHeight) \
 	visit(SmallFontWidth) \
 	visit(SpaceSize) \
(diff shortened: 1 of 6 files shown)


Polymega:
Thanks, Murugo! Really appreciate this. I don't see any harm in keeping it a hidden feature but, if we find it should be toggleable for whatever reason, I'd recommend merging it into `FixAptClockFlashlight` ([see here](https://enhanced.townofsilenthill.com/SH2/config.htm#FixAptClockFlashlight)) and possible renaming the feature to something more generic.

Thank you again!

elishacloud:
Thank you @Murugo!

PR #960, opened 2024-05-04
Remove delay before LOAD and CONTINUE display on the main menu

Murugo:
Adding a patch (`MainMenuInstantLoadOptions`) to show LOAD and/or CONTINUE instantly on the main menu if save data already exists.

The game checks whether save data exists by reading sh2pc.sys before the main menu fades in. This happens over multiple frames to prevent the game from hitching on I/O operations, similar to save/load, etc. For me it usually takes between 8 - 14 frames before LOAD and CONTINUE are drawn on the screen.

This patch loops on the function call to `FUN_00452E40` at instruction `00497C7F` to force the sh2pc.sys check and main menu state change to finish in a single frame. In case this check does not complete on time, the patch caps out at 100 attempts before giving up and continuing to the main menu.

Test build:  [d3d8.dll.zip](https://github.com/elishacloud/Silent-Hill-2-Enhancements/files/15211008/d3d8.dll.zip)

### DIFF
--- a/Common/Settings.h
+++ b/Common/Settings.h
@@ -247,7 +247,6 @@
 	visit(LetterSpacing) \
 	visit(LoadModulesFromMemory) \
 	visit(LockResolution) \
-	visit(MainMenuInstantLoadOptions) \
 	visit(NormalFontHeight) \
 	visit(NormalFontWidth) \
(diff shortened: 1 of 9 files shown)


elishacloud:
Thanks @Murugo!

PR #964, opened 2024-05-14
Move game load screen flash fix to a separate feature

Murugo:
`GameLoadFix` contains a patch that prevents the screen from flashing when loading a save game (594a1e0). This PR moves the patch to its own feature called `GameLoadFlashFix`. This will allow speedrunners to disable `GameLoadFix` without encountering the screen flash bug.

Once again adding this as a hidden feature. Please let me know if this should be added to the config tool.


elishacloud:
Thanks @Murugo!

PR #966, opened 2024-05-19
Launcher updates and configs

mercury501:
Hi!
@elishacloud this is a wip PR, I'll let you know when we're ready to merge!

This branch contains the new developments for the launcher to get it ready for the Speedrun Mode, fixes for the option features to ensure the needed files are present, a check to log a warning if the user has SpeedrunMode enabled, but the dll version isn't speedrun.

Cheers!


elishacloud:
@Murugo, it looks like you added `d3d8.ini` to the repository.  I don't think that is needed since we already have `settings.ini`,which should be the same thing.

elishacloud:
I just merged the latest changes from master in.  It looks like there is still one extra setting in here for `ImprovedWaterFX`.  Once that setting is created the build should pass.

PR #971, opened 2024-05-25
Fix Wood Side Apartments gate lock appearance and commentary

Murugo:
This PR renames the `FixTownWestGateEvent` feature to `FixTownGateEvents` and includes two new patches to fix some inconsistencies:

- The lock on the Wood Side Apartments gate still appears on the gate even after James uses the key and passes through. **Fix**: The lock will now appear on the ground when James is inside the gate.
- When James returns to the Wood Side Apartments gate at night, he will note that the lock is broken, even though the gate is clearly open with the lock on the ground. **Fix**: James will now remark that there is no reason to go back (similar text used elsewhere in the town). This also removes the broken lock SFX when James interacts with the gate.

<details><summary>Before + After</summary>

sh2ee-gateevents-0
sh2ee-gateevents-1

</details>

### DIFF
--- a/Common/Settings.h
+++ b/Common/Settings.h
@@ -64,7 +64,7 @@
 	visit(FixMemoFading, true) \
 	visit(FixMissingWallChunks, true) \
 	visit(FixSaveBGImage, true) \
-	visit(FixTownWestGateEvent, true) \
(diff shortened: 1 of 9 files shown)


elishacloud:
Thanks @Murugo!

PR #986, opened 2024-06-22
Restore flashlight reflections on buildings throughout Silent Hill, and introduce flashlight reflections to some special objects

FrozenFish24:
Huge thanks to @iOrange for making this possible. This is 99% his work (and it looks amazing, might I add).

shop
hospital


FrozenFish24:
Added the addresses for 1.1 and DC and tested that they work. Thanks @AeroWidescreen for doing the legwork in finding them.

elishacloud:
@iOrange, @FrozenFish24, Nice work!

PR #998, opened 2024-08-05
Increase water draw distance in more hallways

Murugo:
Adding more addresses to `IncreaseDrawDistance` to always render water chunks in other hallways. This fixes an issue where water chunks popping in/out is more apparent with brighter screen settings or with specular highlights on the water surface.

Thanks to Aero_ for providing the list of addresses!


elishacloud:
Thanks @Murugo!

PR #1004, opened 2024-08-21
Adjustments to water appearance and draw order

Murugo:
This PR applies several fixes to improve the appearance of water in certain areas of the game.

If `WaterEnhancedRender` is enabled:
- Fix the draw order of water objects and particle effects that appear underwater (blood pools, dust clouds). This resolves a bug where certain effects such as sprays, water splashes, and gun flashes do not render correctly on top of water.
- Prevent blood drops from spawning in rooms with water.
- Disable water UV stretching in the hallway after the Alternate Hotel Kitchen.

Other fixes:
- Adjust water opacity in the Blue Creek Apartments Stairway (first PH boss fight).
- Adjust water brightness in the Alternate Hotel Staircase after the cutscene with Angela.
- Adjust water brightness in the Alternate Hotel Kitchen.

### DIFF
--- a/Common/Settings.h
+++ b/Common/Settings.h
@@ -84,7 +84,6 @@
 	visit(HospitalChaseFix, true) \
 	visit(HotelWaterFix, true) \
 	visit(ImproveStorageSupport, true) \
-	visit(ImprovedWaterFX, false) \
 	visit(IncreaseBlood, true) \
 	visit(IncreaseDrawDistance, true) \
 	visit(LegacyFixGPUAntiAliasing, false) \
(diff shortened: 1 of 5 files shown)


Polymega:
Thanks Murugo!! One thing I noticed:

Can you please add `0xAB, R_HTL_ALT_EMPLOYEE_STAIRS` to the Room ID list for the draw order fix? Even though there's no enemies in this little room, you can still shoot your gun so, without including this room, the muzzle flash/smoke and Hyper Spray will be affected by the shader.

PR #1009, opened 2024-08-24
Disable options if custom strings are missing

Murugo:
This PR disables certain options if required strings are missing in at least one .res file.

- Unset `EnableMasterVolume` if the name and/or description for "Master Volume" are missing, and revert back to "Speaker Config." in Options
- Hide "Render Resolution" in Options > Advanced Options if the name and/or description are missing
- Hide "Health Indicator" in Options > Advanced Options if the name and/or description are missing


elishacloud:
Thanks @Murugo!

PR #1012, opened 2024-08-29
Cancel quick saving 

Murugo:
Small change to `QuickSaveCancelFix` to cover more cases when deciding to cancel an active quick save since these other state changes can have unintended side effects. This fixes the following bugs:

- Quick saving and immediatrly opening the map or inventory screen prevents being able to return to the game via Esc.
- Quick saving right before a cutscene or FMV prevents being able to skip the cutscene.


elishacloud:
Thanks @Murugo!

PR #1030, opened 2024-09-01
Fixed SR settings

mercury501:


elishacloud:
Maybe it is a good idea to remove the code [here](https://github.com/elishacloud/Silent-Hill-2-Enhancements/pull/1030/files#diff-e3724a24155d22093882751c002e2665d21e1f28bda87ebb679955f9380b7d8fL1317-L1321) so that you are not drawing the code twice?

mercury501:
> Maybe it is a good idea to remove the code [here](https://github.com/elishacloud/Silent-Hill-2-Enhancements/pull/1030/files#diff-e3724a24155d22093882751c002e2665d21e1f28bda87ebb679955f9380b7d8fL1317-L1321) so that you are not drawing the overlay twice?

The issue is, the code you mentioned makes it show up during normal gameplay, the code I added makes it show up in obs

Wrappers/d3d8/IDirect3DDevice8.cpp
```diff
@@ -1334,6 +1334,12 @@ HRESULT m_IDirect3DDevice8::DrawScaledSurface()
 		ProxyInterface->SetRenderTarget(pBackBuffer, pStencilBuffer);
 	}
 
+	// Draw Overlays
+	OverlayRef.DrawOverlays(ProxyInterface, Desc.Width * ScaleFactor, Desc.Height * ScaleFactor);
```
elishacloud, line 1338: This concerns me.  If we are sending `DrawOverlays` in a different resolution from the last time we called it `Desc.Width` vs `Desc.Width * ScaleFactor` that causes it to reinitialize the overlay.  Since we are doing this every frame, we are causing it to reinitialize the overlay every frame.  I don't think that is a good idea.

mercury501: Code was removed in the last commit!

elishacloud:
Thanks @mercury501!

PR #1110, opened 2024-11-12
Fix crash when toggling advanced filters from options

Murugo:
This fix resolves a crash caused by enabling or disabling "Advanced Options > Advanced Filters" in certain rooms, e.g. Room 312 in Lakeview Hotel. The new advanced options handler was not allocating enough space on the stack before calling into code that updates advanced filters (`00464E74` - `00464EBB`), causing this code block to corrupt the stack.


elishacloud:
Nice work!  Thanks @Murugo!

PR #1118, opened 2024-11-27
Better bugs

iOrange:
Implemented support for animated gltf models, and using that to replace ugly cockroaches the original game has with way better ones modelled by @Polymega


elishacloud:
Thanks @iOrange!

PR #1126, opened 2024-12-27
Gamma no reshade

iOrange:
Implemented Brightness (gamma ramp) as a simple 3D LUT, so even raw D3d8 device can use it, no ReShade needed


Polymega:
Thank you, iOrange!!

Elisha, our original (ReShade) implementation for restoring the game's brightness selector utilized this shader: 
https://github.com/elishacloud/Silent-Hill-2-Enhancements/blob/master/Resources/LiftGammaGain.fx

I'm not sure if any of our other ReShader shaders require/use `LiftGammaGain` but, if not, that shader file could possibly be removed from our project going forward.

iOrange:
Thanks @Polymega for testing, added small fix to support resolution scale, as well as optimized LUT generation (I'm an idiot at times) so it's smooth now and not lagging when you change brightness level in the menu.

iOrange:
@elishacloud fixed all of the issues and optimized even further.  Tested by @Polymega

elishacloud:
I see a few issues with the latest check-ins here:

1. In the case of scaled resolutions there is an extra, unneeded `CopyRects()` and `DrawPrimitiveUP()`.
2. There is a `DrawPrimitiveUP()` outside the BeginScene()/EndScene() pair, which is not allowed in Direct3D9.

elishacloud:
Ok, I fixed both of the issues.  I reduced the duplication and made sure that all draw calls are between the BeginScene()/EndScene() pair.  Hopefully, I did not introduce a new issue.

elishacloud:
Thanks @iOrange!

iOrange:
Thanks for additional fixes @elishacloud !

PR #1127, opened 2025-01-07
Fixed save number for the info overlay

mercury501:
Changed the save number variable type from int8_t to uint16_t

Found out that after reaching 0xFFFF saves, trying to save again corrupts the save folder.
Quite unlikely, but a little fun fact xD

Waiting for a couple of tests from @Polymega 

Cheers!


mercury501:
Seems all is good, it's ready to be merged!

elishacloud:
Thanks @mercury501!

PR #1184, opened 2025-12-02
+ added patch to replace console model in secret dog room

iOrange:


elishacloud:
@iOrange, it looks like there are some failures here.  See errors here:  https://github.com/elishacloud/Silent-Hill-2-Enhancements/actions/runs/19875738116/job/56962535037?pr=1184

iOrange:
should be good now

iOrange:
Thank **YOU**!

PR #1185, opened 2025-12-07
Play Unused Audio "Laura's Letter"

Sonderhen:
During Laura’s letter scene, the unused audio file “SCENE_45”, is played.
The audio is triggered in specific time intervals that match each corresponding portion of the letter as it appears on screen.

For each different text on the screen, it will play a different audio excerpt.

Tested in all languages

https://github.com/user-attachments/assets/89aa3b80-fe3a-4abe-9894-5f71e49bcbd2


Polymega:
@elishacloud this is good to be merged upon your review/convenience. Thanks and Merry Christmas!

PR #1191, opened 2025-12-29
Support for extending the Eddie and Laura cutscene in the bowling alley

Murugo:
Adding patch `EddieLauraCutscene` in order to use modified .DDS and .ANM files that extend the cutscene between Eddie and Laura in the bowling alley while the screen is fading out.

- Fade the cutscene out at frame 1190 instead of when the cutscene ends.
- Increase the frame count and end frame for animations for Eddie, Laura, and the pizza object to match the number of frames in the .DDS.

Files for testing: [bp_edi_1.zip](https://github.com/user-attachments/files/24362655/bp_edi_1.zip)

https://github.com/user-attachments/assets/775fbc3f-4ad3-4440-979b-a5363ccb56a5


elishacloud:
Thanks @Murugo!

PR #1204, opened 2026-03-08
Add in-memory playback for WAV audio stored in AFS

Sonderhen:
- Uses the existing DirectSound device from the project.
- Creates buffers using the project's wrapper system.
- Registers the buffer so StopWavFile can stop playback correctly.
- Allows bypassing SFX volume through a parameter, since this audio is not intended to behave like a regular SFX.

@elishacloud  Thank you for the suggestions and guidance.

I followed the approach of using the project's existing DirectSound wrapper and buffer system. The only part that required a slightly different solution was the playback itself. In this case it wasn't possible to use PlayWavFile directly, since it expects a WAV file from disk, while this feature reads WAV data from inside the AFS archive and plays specific time ranges from it.

Because of that, I added PlayWavMemory so the audio can be played directly from the PCM data already loaded in memory.

If I misunderstood anything or if there is a better way to integrate this with the existing code, please let me know.

### DIFF
--- a/Common/ScopeGuard.h
+++ b/Common/ScopeGuard.h
@@ -1,0 +1,28 @@
+#pragma once
+
+#define WIN32_LEAN_AND_MEAN
+#include <Windows.h>
+
+struct ScopedCriticalSection
+{
+private:
+    bool enable;
+    CRITICAL_SECTION* cs;
+public:
+    // Constructor enters critical section
+    ScopedCriticalSection(CRITICAL_SECTION* cs, bool setenable = true) : cs(cs), enable(setenable)
+    {
+        if (enable && cs)
+        {
+            EnterCriticalSection(cs);
+        }
+    }
+    // Destructor leaves critical section
+    ~ScopedCriticalSection()
+    {
+        if (enable && cs)
+        {
+            LeaveCriticalSection(cs);
+        }
+    }
+};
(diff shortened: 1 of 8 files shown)


elishacloud:
Any update here?  I hate to hold this PR because of a separate unrelated issue.  If this PR is good I would like to merge it.

Polymega:
Earlier today iOrange whittled down the cause of the audio issue I found and it shouldn't have any affect to Sonderhen's work here.

I'll play test this tomorrow and try to break it then report back.

Polymega:
@Sonderhen thanks for being so patient with me on this. I playtested the revamped feature and, overall, it's totally solid. The only hiccup I found is that you can press "Quick Load" (F9, by default) while the letter is being read, which we don't want to have happen. I believe you addressed this previously but I can't fully remember?

Besides this, everything else works wonderfully. I tried every trick I could think of to mess things up and it all stayed solid.

Sonderhen:
Hi @Polymega thanks for testing! When pressing F9 "Quick Load", does Mary continue reading the letter, or does the audio stop immediately? In PR #1185, the focus regarding Quick Load was on fixing the issue where audio would continue after loading. I don’t think we covered preventing Quick Load while the letter is being read. If you agree, I can try to implement this in another PR, what do you think?

Polymega:
> the focus regarding Quick Load was on fixing the issue where audio would continue after loading

Ah, that's right! My apologies; I couldn't remember the details about that. In that regard, everything is working fine when pressing Quick Load. With that, everything looks completely solid then! Thank you very much again for your great work here, Sonderhen. 👊 

@elishacloud this is good to check-in at your convenience.

elishacloud:
Thanks @Sonderhen for your help here!

PR #1214, opened 2026-05-03
Draw rain as billboarded particles

FrozenFish24:
Replaces the 1 px thick lines used for rain and water droplets throughout the game with resolution-independent particles.

I got a bit confused with the different coordinate systems so I'm not 100% why I needed to invert the up vector on line 137 or if I'm doing something else wrong.

Also I think the DogRoomEnhanced and ClosetCutscene patches may have been broken on 1.1 and DC so I moved the retrieval of the d3d8device to a function that these now all use. ~~I return a void* because including d3d8.h in Patches.h is probably a bad idea?~~ forward declaration is cleaner.

image


elishacloud:
Thanks @FrozenFish24!

PR #1230, opened 2026-08-06
Exclude Heaven's Night neon signage from receiving shadows

FrozenFish24:
Also unify the existing shadow exclusions.

neon


elishacloud:
Thanks @FrozenFish24 !

PR #1234, opened 2026-09-12
Fix movement speed when entering and exiting water

Murugo:
Adding patch `WaterMoveSpeedFix` which adjusts James' movement speed when entering and exiting water in certain flooded areas:

- Strange Area (beneath the historical society): Restore full movement speed when stepping out of the water.
- Hotel Alt Elevator: Slow down movement speed after the elevator floods with water.
- Hotel Alt Employee Stairs (after cutscene with Angela): Restore full movement speed when climbing the stairs out of the water.

Also cleaned up some redundant code across other patches that checks whether certain game flags are set.

### DIFF
--- a/Common/Settings.h
+++ b/Common/Settings.h
@@ -298,6 +298,7 @@
 	visit(SmallFontWidth) \
 	visit(SmokeFogFix) \
 	visit(SpaceSize) \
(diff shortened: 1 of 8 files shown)


elishacloud:
Nice work.  Thanks @Murugo !

PR #1235, opened 2026-09-13
Updates to .GLB model replacements

Murugo:
Adding various updates and bug fixes to model replacements:

- **Maria**: Add patch `MariaCutsceneReplacement` which replaces `dmr.mdl` during the cutscene before the final RPT fight.
- **Apt Room 307**: Add support for versions 1.1 and DC for the closet door.
- **Dog Ending**: Always start the console model animation on the first frame.
- **All**:
  - Skip replacing a model if the corresponding file does not exist in `sh2e\model`.
  - Load models during the screen transition before the cutscene begins to avoid hitching during the cutscene.
  - Sync model animations to the cutscene timer.


elishacloud:
Thanks @Murugo !

PR #1238, opened 2026-09-16
Replace jacket texture for James during noose cutscene

Murugo:
Adding patch `JamesTextureFix` which replaces James' jacket texture during the cutscene where James pulls down a noose (`CS_PS_NOOSE_PULL`). The texture hides parts of the model that clip through James' shoulders when he raises his arms:

image

This assumes the replacement texture has been inserted into `bg\ps\ps113.map` with ID 9980.

(Adding the texture to one of the character models (`chr\jms\*_jms.mdl`) is not viable since the game will only load textures for James once, typically from the low-poly in-game model.)


elishacloud:
Thanks @Murugo !

